Issue #2, opened 2019-01-17
Let user resolve dependency conflicts


accuminium:
Currently, bender reports an error when it encounters conflicting version requirements on a dependency. A common scenario where this happens is when I set the revision of a dependency to a feature branch to make a feature that is not yet released available early in development.

In this case, I would like to be able to resolve the conflict interactively and store my decision in the `Bender.lock` file.

The interaction could be as follows:
```
Requirement `^1.2.3` conflicts with other requirements on dependency `common_pkg`:
- package `my_pkg` requires `new_feature`
- package `other_pkg` requires `^1.2.3`

To resolve this conflict manually, select a revision for `common_pkg` among:
1) `1.2.3`
2) `new_feature`
Enter a number or hit enter to abort: 
```

If stdin is closed, bender should maintain its current behavior.

fabianschuiki:
This has been completed as of 6ae2a71.

PR #3, opened 2019-03-11
Let user resolve dependency conflicts (#2)

accuminium:
Let's get #2 implemented.

- [x] Let the user interactively resolve dependency conflicts as described in #2.
- [x] Impose manually selected revision for that dependency on *all* packages. Currently the user query is repeated for every conflicting package.
- [x] Maintain old behavior if `stdin` is not open.
- [x] Properly handle I/O and parsing problems; return descriptive errors, repeat input, or return the original error  (lots of dirty `unwrap`s at the moment).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ tokio-process = "0.1.5"
 tokio-timer = "0.1"
 
 clap = "2.20.0"
-semver = "0.5.1"
+semver = "0.9.0"
 blake2 = "0.7.0"
 typed-arena = "1.3.0"
 dirs = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,3 +23,5 @@ dirs = "1"
 pathdiff = "0.1"
 
 itertools = "0.8.0"
+
+atty = "0.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,3 +21,5 @@ blake2 = "0.7.0"
 typed-arena = "1.3.0"
 dirs = "1"
 pathdiff = "0.1"
+
+itertools = "0.8.0"
--- a/src/resolver.rs
+++ b/src/resolver.rs
@@ -5,12 +5,12 @@
 
 #![deny(missing_docs)]
 
+use std::collections::{HashMap, HashSet};
 use std::fmt;
 use std::mem;
-use std::collections::{HashMap, HashSet};
 
-use futures::Future;
 use futures::future::join_all;
+use futures::Future;
 use tokio_core::reactor::Core;
 
 extern crate itertools;
@@ -20,9 +20,12 @@ extern crate atty;
 
 use std::io::{self, Write};
 
-use sess::{self, Session, SessionIo, DependencyVersions, DependencyVersion, DependencyRef, DependencyConstraint};
-use error::*;
 use config;
+use error::*;
+use sess::{
+    self, DependencyConstraint, DependencyRef, DependencyVersion, DependencyVersions, Session,
+    SessionIo,
+};
 
 /// A dependency resolver.
 pub struct DependencyResolver<'ctx> {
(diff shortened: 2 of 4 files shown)


accuminium:
@fabianschuiki Feel free to comment and improve this first implementation; I am sure my code is not as rusty as it should be. ;-)

Cargo.toml
```diff
@@ -16,8 +16,10 @@ tokio-process = "0.1.5"
 tokio-timer = "0.1"
 
 clap = "2.20.0"
-semver = "0.5.1"
+semver = "0.9.0"
```
fabianschuiki, line 19: Good idea to keep these things up to date.

src/resolver.rs
```diff
@@ -349,23 +368,57 @@ impl<'ctx> DependencyResolver<'ctx> {
...
+                    match self.req_indices(name, decision, src) {
+                        Ok(o) => match o {
+                            Some(v) => Ok(v),
+                            None => Err(Error::new("something"))
+                        },
+                        Err(e) => Err(e)
                     }
-                    return Err(Error::new(msg));
+                } else {
+                    Ok(is_ids)
```
fabianschuiki, line 476: Pretty elegant! I like the ability for the user to resolve conflicts. As you say in the PR, it would be great if we could wrap the line-reading in a loop that keeps asking the user for input as long as the input is invalid and the user hasn't hit enter.

fabianschuiki:
Generally I think we should get this merged quickly such that we have a working version of dependency resolution with user feedback. I want to rewrite the dependency resolver at some point to allow for proper incremental unlocking of dependencies, but that can wait and is not yet a high priority. I'd rather keep bender useful and get improvements in quickly, instead of waiting forever for some internal rewrite :wink:.

Any thoughts?

Cargo.toml
andreaskurth: Yes, and it's a requirement for `Eq` and `Hash`, see 2e8885c9c6d2c0e69a07a3faef2b886893ed50e8.

src/resolver.rs
andreaskurth: Done, see the new commits.

accuminium:
I totally agree we should not write for the rewrite of the dependency resolver. I think the feature in this PR is now in a pretty good state, with the exception of the remaining TODO item
> Impose manually selected revision for that dependency on all packages. Currently the user query is repeated for every conflicting package.

I would argue that this is necessary to make this feature really usable. Currently the same query is repeated many times (as in 10x) if a dependency conflict occurs for a common package, which is very annoying.

As far as I see, we should apply the user-selected dependency to all other packages that are in `all_cons` https://github.com/fabianschuiki/bender/blob/56cbe580e6d45d730c149ef23bcd6f972b377fc0/src/resolver.rs#L382
but we only know the package names, not the actual `DependencySource`. As you are more familiar than me with the architecture, do you see an easy way to bring the actual `DependencySource`s for manipulation to this place?

Cargo.toml
fabianschuiki: Ah good catch!

fabianschuiki:
We could add a hash table of user-resolved conflicts after the resolution table:
https://github.com/fabianschuiki/bender/blob/56cbe580e6d45d730c149ef23bcd6f972b377fc0/src/resolver.rs#L32
Something like:
```rust
HashMap<&'ctx str, usize>
```

Then whenever the user makes a decision, we add that to the table. Before asking the user for a decision, we could check back with the hash map to see if we have already asked before.

Would that cover all use cases?

fabianschuiki:
@accuminium The user's decisions are now being cached. Shall we merge?

accuminium:
Very nice, thanks! Just tested, works like a charm.

:+1: for merge

accuminium:
Very nice, thanks! Just tested, works like a charm.

:+1: for merge

PR #6, opened 2019-04-26
sources: Enable flattened and pretty print source listings

zarubaf:


fabianschuiki:
Very nice, thanks a lot! Merging.

PR #7, opened 2019-05-07
script: Fix two bugs with target defines

accuminium:


fabianschuiki:
Looks good to me, thanks!

PR #9, opened 2019-07-24
Automatically deploy release binaries from Travis

accuminium:
With this patch, Travis build release binaries for Linux and two recent CentOS versions (to precisely match the target libc) for every tag and attaches them as asset to a release on GitHub, see [this example](https://github.com/fabianschuiki/bender/releases/tag/ci-test).

### DIFF
--- a/.ci/if_deploy_build.sh
+++ b/.ci/if_deploy_build.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 
-if [ "$TRAVIS_RUST_VERSION" = "1.35.0" ]; then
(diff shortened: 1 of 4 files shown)


fabianschuiki:
This looks excellent, many thanks! Once we have fixed the bender build under stable I guess we can revert the rustc version "1.35.0" check back to "stable"? Merging.

accuminium:
Yes, sure! If you want deploy builds to be compiled with `stable`, you need to change `1.35.0` in https://github.com/fabianschuiki/bender/blob/83ab691d02f03046451babe9ef89215339c73298/.ci/if_deploy_build.sh#L3 to `stable`. If this condition is not met by any build, nothing will be deployed.

PR #10, opened 2019-07-24
script: Capitalize target defines again

zarubaf:
I think that fixes a regression introduced with the new feature-set. Feel free to deny or request changes.


fabianschuiki:
You are absolutely right. Many thanks for the fix!

PR #12, opened 2019-07-29
script: Define format-specific targets even if no target is defined

accuminium:
Emit format-specific target defines even if no target is specified.

Previously, `bender script <format>` would emit target-specific (e.g., `'TARGET_SYNTHESIS`) `define`s only if `-t <some target>` was additionally specified.


fabianschuiki:
Cool thanks!

PR #14, opened 2019-07-29
Add `script vivado`

accuminium:
This adds an initial implementation for `bender script vivado`, which outputs commands to add all source files, set all include directories for all files, and set all defines for all files.

This is a very coarse-grained method, but in Vivado the only way to set include directories and defines on a *subset* of files would be to define multiple filesets, which leads to out-of-context synthesis in Vivado, which is not always desirable. We may later add an option to `script vivado` that allows to do this.

I validated the resulting commands in Vivado 2018.3.

### DIFF
--- a/src/cmd/script.rs
+++ b/src/cmd/script.rs
@@ -28,7 +28,7 @@ pub fn new<'a, 'b>() -> App<'a, 'b> {
             Arg::with_name("format")
                 .help("Format of the generated script")
                 .required(true)
-                .possible_values(&["vsim", "synopsys"]),
+                .possible_values(&["vsim", "synopsys", "vivado"]),
         )
         .arg(
             Arg::with_name("vcom-arg")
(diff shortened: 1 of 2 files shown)


src/cmd/script.rs
```diff
@@ -92,10 +94,10 @@ pub fn run(sess: &Session, matches: &ArgMatches) -> Result<()> {
 /// Files with the same category that appear after each other will be kept in
 /// the same source group. Files with different cateogries are split into
 /// separate groups.
-fn separate_files_in_group<F1, F2, T>(mut src: SourceGroup, categorize: F1, consume: F2)
+fn separate_files_in_group<F1, F2, T>(mut src: SourceGroup, categorize: F1, mut consume: F2)
 where
     F1: Fn(&SourceFile) -> Option<T>,
-    F2: Fn(&SourceGroup, T, Vec<SourceFile>),
+    F2: FnMut(&SourceGroup, T, Vec<SourceFile>),
```
andreaskurth, line 100: @fabianschuiki: I had to change this and line 97 to be able to manipulate `defines` from inside the closure on line 333. Was this the proper way to do so?

(Line numbers referring to post-patch file.)

fabianschuiki: Yes this is perfect!

fabianschuiki:
Ready for merging?

accuminium:
I just added a Changelog entry. Now it's ready!

PR #15, opened 2019-08-12
WIP: Add `script xcelium`

bluewww:
Generates a flat list of command line parameters which can be fed to `xcelium` using the `-f` switch. Normally this contains a list of files in the correct order (with respect to dependencies) and defines.
 
TODO
* [ ] fix warnings
* [ ] figure out how and if vhdl files need different handling
* [ ] run tests


accuminium:
Thanks for your work! Please let us know when you need feedback or help.

bluewww:
I'm not convinced of this approach anymore.

PR #16, opened 2019-08-15
script: Add `--abort-on-error` flag

accuminium:
This flag causes compilation/analysis scripts to abort on errors that can be caught with Tcl's `catch`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ before_script:
 script:
   - cargo build
   - cargo test --all
-  - cargo fmt -- --check
+  - "[ $TRAVIS_RUST_VERSION != stable ] || cargo fmt -- --check"
   - tests/run_all.sh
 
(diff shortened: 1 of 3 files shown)


accuminium:
I think it is useful to have an option that lets analysis/compilation abort if errors are encountered.

@fabianschuiki: What do you think? Should this even be the default behavior and should we instead have an `--ignore-errors` switch?

fabianschuiki:
Any thoughts on how we should proceed with this?

accuminium:
I forgot that we never merged this, sorry. I still think it's very useful to abort on first error when possible. I rebased on current `master` and enabled aborting by default, as you suggested.

fabianschuiki:
Perfect thanks! I quickly ran `cargo fmt` over it to make CI pass. Then I'll merge and release.

PR #17, opened 2019-08-23
Add repository information to Cargo.toml

dalance:
I found this project through crates.io. But there is not link to this repository.
I think the link is useful.


accuminium:
Thanks!

PR #20, opened 2020-02-12
Improve `script vivado`

accuminium:


src/cmd/script.rs
```diff
@@ -33,16 +33,36 @@ pub fn new<'a, 'b>() -> App<'a, 'b> {
...
         )
         .arg(
             Arg::with_name("vlog-arg")
                 .long("vlog-arg")
-                .help("Pass an argument to vlog calls")
+                .help("Pass an argument to vlog calls (vsim only)")
                 .takes_value(true)
-                .multiple(true),
+                .multiple(true)
+        )
```
fabianschuiki, line 46: This is an excellent change.

fabianschuiki:
Is this ready to merge from your side? Also, I would suggest we release a new version which contains these changes. What do you think?

accuminium:
Great, thanks! Yes, ready for merge and release from my side as well.

PR #30, opened 2020-07-06
common compilation for Riviera-PRO

dawidzim:
Hey!
In this pull request I added support for common compilation in Riviera-PRO. This compilation mode will be used as default and to change mode we need to type `--compilation_mode=separate`

### DIFF
--- a/src/cmd/script.rs
+++ b/src/cmd/script.rs
@@ -104,6 +104,18 @@ pub fn new<'a, 'b>() -> App<'a, 'b> {
                 .long("no-abort-on-error")
                 .help("Do not abort analysis/compilation on first caught error (only for programs that support early aborting)")
         )
+        .arg(
+            Arg::with_name("compilation_mode")
+                .long("compilation-mode")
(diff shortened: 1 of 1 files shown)


src/cmd/script.rs
```diff
@@ -104,6 +104,18 @@ pub fn new<'a, 'b>() -> App<'a, 'b> {
                 .long("no-abort-on-error")
                 .help("Do not abort analysis/compilation on first caught error (only for programs that support early aborting)")
         )
+        .arg(
+            Arg::with_name("compilation_mode")
+                .long("compilation_mode")
```
fabianschuiki: How about changing the long name to `compilation-mode` to stay in line with the other multi-word switches?

dawidzim: Ok, I changed long name as you suggest

dawidzim:
So I used `cargo fmt` and I combined commits to one and I think now is fine

fabianschuiki:
Perfect, thanks! Merging.

PR #31, opened 2020-07-17
update readme for Riviera

dawidzim:
Hey!
In this pull request I added information about Riviera-PRO in readme. Should I create also `bender-riviera plugin` like is in other tools?


fabianschuiki:
Excellent, thanks! No need to create a plugin, since you have added the Riviera support directly to the Rust source base. The plugins are just there to hook arbitrary scripts into the bender flow without having to touch the rust code base.

PR #33, opened 2020-10-23
cli: Also look for Bender.local recursively

zarubaf:
Bender would only check for a Bender.local in the root of the project
but fail to find any other `Bender.local`s on different hierarchy
levels. It now looks for `Bender.local` and `.bender.yml`.


fabianschuiki (approved):
LGTM, thanks! Merging.

PR #34, opened 2020-11-04
Add command line argument for defines

SamuelRiedel:
Add a flag (`--defines`) which allows giving additional defines through the command line

### DIFF
--- a/src/cmd/script.rs
+++ b/src/cmd/script.rs
@@ -302,13 +302,13 @@ fn tcl_catch_postfix() -> &'static str {
     return "}]} {return 1}";
 }
 
-fn add_defines_from_matches(defines: &mut Vec<(String, Option<&str>)>, matches: &ArgMatches) {
+fn add_defines_from_matches(defines: &mut Vec<(String, Option<String>)>, matches: &ArgMatches) {
     if let Some(d) = matches.values_of("define") {
         defines.extend(
             d.map(|t| {
                 let mut parts = t.splitn(2, "=");
                 let name = parts.next().unwrap().trim(); // split always has at least one element
-                let value = parts.next().map(|v| v.trim());
+                let value = parts.next().map(|v| v.trim().to_string());
                 (name.to_string(), value)
             })
         );
@@ -343,8 +343,8 @@ fn emit_vsim_tcl(
                         if let Some(args) = matches.values_of("vlog-arg") {
                             lines.extend(args.map(Into::into));
                         }
-                        let mut defines: Vec<(String, Option<&str>)> = vec![];
-                        defines.extend(src.defines.iter().map(|(k, &v)| (k.to_string(), v)));
+                        let mut defines: Vec<(String, Option<String>)> = vec![];
+                        defines.extend(src.defines.iter().map(|(k, &v)| (k.to_string(), v.map(String::from))));
                         defines.extend(
                             targets
                                 .iter()
@@ -356,7 +356,7 @@ fn emit_vsim_tcl(
                             let mut s = format!("+define+{}", k.to_uppercase());
                             if let Some(v) = v {
                                 s.push('=');
-                                s.push_str(v);
+                                s.push_str(&v);
                             }
                             lines.push(s);
                         }
@@ -422,8 +422,8 @@ fn emit_vcs_sh(
                         if let Some(args) = matches.values_of("vlog-arg") {
                             lines.extend(args.map(Into::into));
                         }
-                        let mut defines: Vec<(String, Option<&str>)> = vec![];
-                        defines.extend(src.defines.iter().map(|(k, &v)| (k.to_string(), v)));
+                        let mut defines: Vec<(String, Option<String>)> = vec![];
+                        defines.extend(src.defines.iter().map(|(k, &v)| (k.to_string(), v.map(String::from))));
                         defines.extend(
                             targets
                                 .iter()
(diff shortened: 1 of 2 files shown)


src/cmd/script.rs
```diff
@@ -864,6 +871,13 @@ fn emit_vivado_tcl(
             .iter()
             .map(|t| (format!("TARGET_{}", t.to_uppercase()), None)),
     );
+    if let Some(define) = matches.values_of("defines") {
+        defines.extend(
+            matches.values_of("defines")
+            .unwrap()
+            .map(|t| (t.to_string(), None))
```
fabianschuiki:
```suggestion
            define.map(|t| (t.to_string(), None))
```
should work here as well.

src/cmd/script.rs
```diff
@@ -47,6 +47,13 @@ pub fn new<'a, 'b>() -> App<'a, 'b> {
                 .long("relative-path")
                 .help("Use relative paths (flist generation only)"),
         )
+        .arg(
+            Arg::with_name("defines")
+                .long("defines")
+                .help("Pass an additional define to all source files")
+                .takes_value(true)
+                .multiple(true),
+        )
```
fabianschuiki: Very nice idea! Also setting `.multiple(true)` is smart, because then you can pass things like `--defines A=0 B=1 C=2` if there is no ambiguity with other options. Small nitpick: How about calling the long option `--define` (singular), and adding `.short("D")` such that you can also write `-D A=0` (or even `-DA=0` thanks to clap).

src/cmd/script.rs
```diff
@@ -864,6 +871,13 @@ fn emit_vivado_tcl(
             .iter()
             .map(|t| (format!("TARGET_{}", t.to_uppercase()), None)),
     );
+    if let Some(define) = matches.values_of("defines") {
+        defines.extend(
+            matches.values_of("defines")
+            .unwrap()
+            .map(|t| (t.to_string(), None))
+        );
+    }
```
fabianschuiki: Since as you say we'll need this for every script format, I would recommend outlining this code into a separate function, e.g. `fn add_defines_from_matches(defines: &mut Vec<(...)>, matches: &...)`. Then you can just call that function like `add_defines_from_matches(&mut defines, matches)` at this point. And add the call to all the other `emit_*` functions until we come up with a better scheme to streamline things.

src/cmd/script.rs
fabianschuiki: This does not support assigning a value to the define at the moment. I think adding this functionality should be fairly straightforward:
```suggestion
            .map(|t| {
                let mut parts = t.splitn(2, "=");
                let name = parts.next().unwrap().trim(); // split always has at least one element
                let value = parts.next().map(|v| v.trim());
                (name.to_string(), value.to_string())
            })
```
The `splitn` gives you an iterator over the parts of the string separated by "=", and at most 2 (with the second entry containing the remainder of the string). The `trim` then cuts away any potential leading and trailing whitespaces.

fabianschuiki (changes requested):
A very nice addition. See some inline comments :smiley: !

src/cmd/script.rs
SamuelRiedel: Yes, I think adding a `-D` option is a no-brainer.

src/cmd/script.rs
SamuelRiedel: It allowed assigning a value as the first entry of the tuple was already the whole string. That is, with `--define A=0`, the tuple became `("A=0", None)`. But I agree this is cleaner, especially for targets that don't use a equal sign for their assignment (if those targets exist). I also allows for stuff like this: `--define "A = 0"`.

Or did you mean something different with supporting to assign values?

fabianschuiki: Exactly what I meant :+1:. Completely overlooked that for some tools that might still work if they use an `=` for assignment as well. Better to go with the split option though, since the code in the script emission might rely on having a clear separation between defined name and value.

src/cmd/script.rs
SamuelRiedel: I implemented this, but have some questions.
1. How about the lifetime of the variables? I added the proposals the compiler gave me, but I have no clue if this makes sense.
2. Vivado seems to be the only target currently not specifying the type of the `defines` variable and the only target that uses a slightly different format... Aligning the format with other targets breaks the current implementation of parsing the source's defines. See: https://github.com/fabianschuiki/bender/pull/34/commits/32928efd6aaeeeadd065b3b4b946175112dfa790. The problem is that I would assign a reference which is only valid inside the closure. Moving this part outside the closure is probably going to break someone's flow. So, is there an easy way to convert the `defines` from `Vec<(String, Option<String>)>` to `Vec<(String, Option<&str>)>` after this closure?

src/cmd/script.rs
SamuelRiedel: Done.

src/cmd/script.rs
fabianschuiki: Yeah that's a bit annoying; the implementation of `separate_files_in_group` is not ideal, because it assumes ownership of `src`, which gets dropped after the call. Therefore you cannot add references from that to `defines`, which will exist after `src` was freed. Ideally `separate_files_in_group` would take `&src` instead of `src`, but let's postpone that change for the sake of getting things done.

As an alternative, I would suggest making all defines `Vec<(String, Option<String>)>`, and adding `.map(String::from)` where the compiler complains about `&str` vs `String`.

SamuelRiedel: Ok, I changed the format and it seems to be working. :tada:

fabianschuiki (approved):
Looks good to me! If you feel like removing the "WIP:" prefix we can merge right away :+1:

SamuelRiedel:
Thanks a lot for merging and your help!

fabianschuiki:
Thanks for the great and swift contribution! :tada:

PR #41, opened 2021-01-05
Bugfix: plugins work for scripts in the root repository

micprog:
While attempting to use plugin scripts with bender, I noticed that plugins defined in the main repository's bender file would not be found. I believe this fixes the issue, albeit not very cleanly (first time using rust).


fabianschuiki:
Looks good in general, thanks a lot! Can you quickly run `cargo fmt` on your machine such that the CI build passes? Once it passes we can merge :+1:

fabianschuiki (approved):

fabianschuiki:
Awesome thanks! Merging.

PR #42, opened 2021-01-12
Force git fetch on unsatisfied requirements to catch newly added tags

micprog:
When updating the version tag in a Bender.yml file to a newly created github tag, bender complains that it cannot satisfy the requirement, likely as it hasn't fetched from the github repository since the tag was added. This should force one fetch of the repo if the requirement cannot be satisfied, leaving all other behavior identical.

I am not that familiar with rust, please let me know if modifications are needed.


fabianschuiki (approved):
This looks very nice, thanks a lot for the contribution!

PR #43, opened 2021-01-20
Added `clone` and `parents` command for improved workflow

micprog:
Added `clone` command to copy a git dependency to a workable location, adding a path override to Bender.local.
Added `parents` command to list packages calling a specified package.


fabianschuiki:
This looks very good to me. Thanks for the awesome effort! Merging.

PR #44, opened 2021-01-21
Fix tests on recent Git versions

accuminium:
Recent versions of Git default to `main` as default branch name.  This causes the `iss5` test to fail.  This PR fixes this in a backwards-compatible way.

Additionally, `rev` in the `iss2` test is updated to use `main` as default branch name.

### DIFF
--- a/tests/iss2_empty_dep_test.sh
+++ b/tests/iss2_empty_dep_test.sh
@@ -20,7 +20,7 @@ package:
   name: bar
 
 dependencies:
-  foo: { git: \"file://$DIR/foo\", rev: master }
+  foo: { git: \"file://$DIR/foo\", rev: main }
 " > Bender.yml
 if $BENDER path foo &> log; then # this fails according to issue #2
 	cat log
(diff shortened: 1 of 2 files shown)


fabianschuiki (approved):
Nice change, thanks!

fabianschuiki:
Seems to break on Travis CI and older Git versions. I remember there being an option to manually set the name of the initial branch. Maybe we can just use that to enforce a specific branch name, like "dummy", so we don't have to check what Git picked?

accuminium:
> Seems to break on Travis CI and older Git versions.

Ah, the joy of outdated software :-) Can't believe they are running Git older than 2.22 (which was released more than 1.5 years ago).

> I remember there being an option to manually set the name of the initial branch. Maybe we can just use that to enforce a specific branch name, like "dummy", so we don't have to check what Git picked?

Yes, there is `init.defaultBranch`.. but that was introduced in Git 2.28 :p

Anyway, it should be fixed now; let's wait for CI.

fabianschuiki:
> Yes, there is `init.defaultBranch`.. but that was introduced in Git 2.28 :p

Haha, classic dependency hell. Good to see it pass, thanks a lot! Merging :+1:

Issue #50, opened 2021-07-28
Slow 'bender update'


FrancescoConti:
`bender update` seems to be very slow on CentOS machines (for a benderized PULP cluster). Specifically, it seems to hang for 2-3 minutes without printing anything just after it clones all repos.
Any idea about possible causes / workarounds / solutions? Unfortunately, I could not really dig into the bowels of the issue... but it doesn't look like an expected behavior.

accuminium:
Thanks for reporting this issue, @FrancescoConti! This should be fixed in #51, which has been released with v0.23.0.  Please try this or a later version and let us know if the problem persists.

PR #51, opened 2021-09-09
Performance

micprog:
Improve `bender update` performance for large projects with multiple git repositories, as referenced in #50.

Changes:
- resolver only uses one Core and SessionIo that is passed to subfunctions
- git_versions call uses a HashMap to fetch versions instead of repeatedly executing git calls in case multiple repositories reference a single dependency


fabianschuiki (approved):
This looks good to me! Thanks for taking the time and tackling this :+1:

Issue #52, opened 2021-09-14
Too many open files (os error 24)


micprog:
In large repositories, an error can occur on `bender update`: 
```
error: Failed to spawn child process. Too many open files (os error 24)
```
A workaround to mitigate this issue is to increase the allowable file limit for the system by increasing `ulimit -n`, for example to 4096 by running `ulimit -n 4096`. While this currently causes the issue to go away, it doesn't seem like the ultimate solution.

micprog:
Should be fixed by #65. Very large projects may still cause issues, however a hint is given.

PR #53, opened 2021-09-15
Fixes for #52 and --fetch and --local features for update

micprog:
### Added
- Added hint for workaround in issue #52
- Added warning for mismatch of dependency name and name in package
- Added `-fetch` argument to `bender update` to force refetch of git dependencies from their remotes
- Added global `--local` argument to disable remote accesses of git commands, e.g. for air-gapped computers

### Changed
- changed method to get git commit hash from tag to use `git show-ref --dereference` instead of individually calling `git rev-parse --verify HASH^{commit}`, helps with #52
- Update links to `pulp-platform`
- Add correct base path to path dependencies linked within git dependencies


micprog:
Replaced by individual PRs: #63, #65, #66, #67, #69

PR #55, opened 2021-10-03
Add script for Mentor Precision

WRoenninger:
This adds the feature to generate scripts for adding files to Mentor's Precision RTL synthesis tool.

Also this is the first time that I wrote something rust. So feel free to correct this if I made some obvious mistakes / broke the cargo dependencies.


micprog:
Hi @WRoenninger, thanks for the PR! Overall looks good, a few minor things:
- could you run `cargo fmt` to fix formatting?
- Looking at the CI, it seems the Cargo.lock reshuffling exposes an issue with the updated bitflags dependency (used in clap) in older rust versions <1.46.0. It would probably make sense to update the minimum supported rust version in `.github/workflows/ci.yml` to 1.46.0. 
- I am not very familiar with Mentor Precision RTL, however I believe it is used for fpga synthesis. Would it make sense to add the fpga target? Alternatively the user would be responsible for adding this target if needed.
- As I am not very familiar with Mentor Precision, I have not checked if the generated scripts work as expected, however generating scripts worked for me, so if you checked the syntax it should be ok.

WRoenninger:
Hi @micprog, added the suggestions to the PR.
* Executed `cargo fmt`
* Updated rust min version to 1.46.0
* Added `fpga` as default target.
As for the syntax of the generated script, was using a patched `bender` with these changes locally already in my workflow.

On that note, more perhaps of a feature request to be handled separately. I am working with some VHDL sources currently and some codebases use different compilation library names instead of the default `work`. I think there is not really a way of specifying a compilation library per file group. How big would the effort probably be to add such a feature?

micprog:
Thanks @WRoenninger ! LGTM now.

Regarding the use of different libraries, you may want to check out a WIP branch `dep_info`, where you can generate scripts for individual packages with or without its dependencies, as well as excluding individual packages. If I'm not mistaken you can then map the corresponding dependencies to different libraries, e.g. in a Makefile calling bender. If different behavior is desired, I suggest opening an issue to discuss this further.

micprog (approved):
Thanks for the changes, merging

PR #57, opened 2021-10-29
Add error on cyclic dependencies to avoid infinite loop

micprog:
Currently, if a dependency calls itself (or a different dependency subsequently calling itself, or further chains), this creates a dependency loop and results in an infinite loop within bender. This PR returns an error if cyclical dependencies are detected, with hints where to find the issue.

### DIFF
--- a/src/resolver.rs
+++ b/src/resolver.rs
@@ -218,6 +218,13 @@ impl<'ctx> DependencyResolver<'ctx> {
                     name, manifest.package.name
                 )));
             }
+            if name == manifest.package.name {
+                return Err(Error::new(format!(
+                    "Please ensure no packages with same name as calling package\n\
+                    \tCurrently {} is called in {}",
+                    name, manifest.package.name
+                )));
+            }
             self.register_dependency(name, id, versions[&id].clone());
         }
         Ok(())
--- a/src/resolver.rs
+++ b/src/resolver.rs
@@ -211,6 +211,13 @@ impl<'ctx> DependencyResolver<'ctx> {
 
         // Register the versions.
         for (name, id) in names {
+            if name == self.sess.manifest.package.name {
+                return Err(Error::new(format!(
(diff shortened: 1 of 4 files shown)


src/sess.rs
```diff
@@ -190,6 +191,20 @@ impl<'sess, 'ctx: 'sess> Session<'ctx> {
                             pending.insert(dep_id);
                         }
                     }
+                    if ranks[&id] > 2 * graph.len() {
```
andreaskurth, line 196: Add a comment describing this condition. For example:
```suggestion
                    //  Limit rank to two times graph length, which is sufficient except if there is
                    // a cyclic dependency.
                    if ranks[&id] > 2 * graph.len() {
```

src/resolver.rs
```diff
@@ -211,6 +211,13 @@ impl<'ctx> DependencyResolver<'ctx> {
 
         // Register the versions.
         for (name, id) in names {
+            if name == self.sess.manifest.package.name {
```
andreaskurth, line 214:
```suggestion
            if name == manifest.package.name {
```
Since `manifest` is a function argument, that should be used instead of the `manifest` in `sess`, right? The `register_dependencies_in_manifest` function is not always called with `manifest == self.sess.manifest`.

accuminium (changes requested):
Thanks! LGTM overall, I just have two suggestions

src/resolver.rs
micprog: This is true, however the behavior is different, as in one case the top package name is compared with all dependencies called throughout the dependency tree, while in the other case it is only checked if a package does not depend on itself. I have added the second test with the new error case as well.

andreaskurth: I see, thanks for elaborating!

accuminium (approved):
LGTM now :+1:

PR #59, opened 2021-11-25
Wrap defines in quotes for the VCS's shell script

SamuelRiedel:
Since VCS generates a shell script, we have to wrap the defines in quotes for this case to allow defines like `32'h80000000`. Otherwise, the `'` breaks the shell script.


micprog (approved):
LGTM, merging

PR #63, opened 2021-12-07
Update links in README.md to PULP-platform

micprog:


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Changed
+- Update links in `README.md` to PULP-platform
```
andreaskurth:
```suggestion
```

I suggest to only track changes that directly affect users of a program or library in the changelog. Changes to tests, CI, documentation, and so on should not be noted in the changelog IMO, because they distract users from changes that directly affect them.

src/git.rs
```diff
@@ -153,7 +153,7 @@ impl<'git, 'io, 'sess: 'io, 'ctx: 'sess> Git<'io, 'sess, 'ctx> {
 
                                 // Parse the ref. This is needed since the ref for an annotated
                                 // tag points to the hash of the tag itself, rather than the
-                                // underlying commit. By callign `git rev-parse` with the ref
+                                // underlying commit. By calling `git rev-parse` with the ref
```
andreaskurth, line 156: Very strictly speaking, this typo fix has nothing to do with the "Update links in README" PR. But since it is part of a separate commit, I would not insist but suggest to FF-merge this branch (I think GitHub calls this *Rebase and Merge*).

accuminium (changes requested):
LGTM overall, thanks! Only two small suggestions

accuminium (approved):

PR #67, opened 2021-12-07
Add warning for name mismatch of dependency and package

micprog:
A naming mismatch can currently cause issues, e.g. when using the parents command to identify packages. May be changed to an error in the future to enforce correct behavior.


accuminium (approved):
LGTM, thanks!

PR #69, opened 2021-12-07
Add arguments to manage git fetching

micprog:
- `bender update --fetch` forces a re-fetch of the git dependencies
- `bender --local` disables fetching of remote repositories to avoid internet access

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -33,7 +33,16 @@ pub fn main() -> Result<()> {
                 .global(true)
                 .help("Sets a custom root working directory"),
         )
-        .subcommand(SubCommand::with_name("update").about("Update the dependencies"))
+        .subcommand(
+            SubCommand::with_name("update")
+                .about("Update the dependencies")
+                .arg(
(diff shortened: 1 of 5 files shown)


src/cli.rs
```diff
@@ -33,7 +33,23 @@ pub fn main() -> Result<()> {
                 .global(true)
                 .help("Sets a custom root working directory"),
         )
-        .subcommand(SubCommand::with_name("update").about("Update the dependencies"))
+        .arg(
+            Arg::with_name("local")
+                .short("l")
```
andreaskurth:
```suggestion
```
I would prefer not to introduce a global short flag `-l` as alias for `--local`.  `-l` could abbreviate `--list`, `--link`, or similar, so I don't consider it intuitive.

If you agree, please update the mentions of the flag as well.

accuminium (changes requested):
Useful change, thanks! I have one suggestion

accuminium (approved):

PR #70, opened 2021-12-07
Add flags to get sources and scripts for a subset of packages

micprog:
Allows the user to individually specify which packages to include, exclude, and to remove all dependencies from the sources and scripts command. Default behavior (i.e. without the new flags) remains unchanged.


src/src.rs
```diff
@@ -101,6 +103,108 @@ impl<'ctx> SourceGroup<'ctx> {
         )
     }
 
+    /// recursively get dependency names
```
andreaskurth:
```suggestion
    /// Recursively get dependency names.
```

src/src.rs
```diff
@@ -101,6 +103,108 @@ impl<'ctx> SourceGroup<'ctx> {
...
+                    result.extend(group.get_deps(&result, excludes));
+                }
+                _ => {}
+            }
+        }
+
+        result
+    }
+
+    /// Get list of packages bassed on constraints
```
andreaskurth:
```suggestion
    /// Get list of packages based on constraints.
```

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 - Add `--fetch/-f` argument to `bender update` to force re-fetch of git dependencies from their remotes
 - Add global `--local` argument to disable remote accesses of git commands, e.g. for air-gapped computers
 - Add `precision` format to the script command.
+- Add flags to get sources and scripts for a subset of packages
```
andreaskurth:
```suggestion
- Extend the `sources` and `scripts` commands with three flags to work with a subset of packages (`-p`/`--package`, `-e`/`--exclude`, and `-n`/`--no-deps`; see ReadMe or command help for the description).
```

README.md
```diff
@@ -367,6 +367,14 @@ The manifest is recursive by default; meaning that dependencies and groups are n
 
 To enable specific targets, use the `-t`/`--target` option.
 
+To get the sources for a subset of packages, exclude specific packages and their dependencies, or exclude all dependencies, the following flags exist:
+
+- `-p`/`--package`: Specify package to show sources for.
+- `-e`/`--exclude`: Specify package to exclude from sources.
+- `-n`/`--no_deps`: Exclude all dependencies, i.e. only top level or specified package(s).
```
andreaskurth:
```suggestion
- `-n`/`--no-deps`: Exclude all dependencies, i.e. only top level or specified package(s).
```
This should be a dash to stay consistent with existing flags (e.g., `--complation-mode`, `--no-abort-on-error`).

src/cmd/script.rs
```diff
@@ -126,13 +126,48 @@ pub fn new<'a, 'b>() -> App<'a, 'b> {
...
+                .long("package")
+                .help("Specify package to show sources for.")
+                .takes_value(true)
+                .multiple(true)
+                .number_of_values(1),
+        )
+        .arg(
+            Arg::with_name("no_deps")
+                .short("n")
+                .long("no_deps")
```
andreaskurth:
```suggestion
                .long("no-deps")
```

src/cmd/sources.rs
```diff
@@ -33,20 +34,55 @@ pub fn new<'a, 'b>() -> App<'a, 'b> {
...
+                .long("package")
+                .help("Specify package to show sources for.")
+                .takes_value(true)
+                .multiple(true)
+                .number_of_values(1),
+        )
+        .arg(
+            Arg::with_name("no_deps")
+                .short("n")
+                .long("no_deps")
```
andreaskurth:
```suggestion
                .long("no-deps")
```

src/cmd/script.rs
```diff
@@ -126,13 +126,48 @@ pub fn new<'a, 'b>() -> App<'a, 'b> {
                     "common",
                 ])
         )
+        .arg(
+            Arg::with_name("package")
+                .short("p")
+                .long("package")
+                .help("Specify package to show sources for.")
```
andreaskurth:
```suggestion
                .help("Specify package to show sources for")
```
No other help string for `script` ends with a period (absolute nitpicking, just struck my eye when I ran `script -h` :slightly_smiling_face:).

src/cmd/script.rs
```diff
@@ -126,13 +126,48 @@ pub fn new<'a, 'b>() -> App<'a, 'b> {
...
+                .takes_value(true)
+                .multiple(true)
+                .number_of_values(1),
+        )
+        .arg(
+            Arg::with_name("no_deps")
+                .short("n")
+                .long("no_deps")
+                .help("Exclude all dependencies, i.e. only top level or specified package(s)."),
```
andreaskurth:
```suggestion
                .help("Exclude all dependencies, i.e. only top level or specified package(s)"),
```

src/cmd/script.rs
```diff
@@ -126,13 +126,48 @@ pub fn new<'a, 'b>() -> App<'a, 'b> {
...
+        )
+        .arg(
+            Arg::with_name("exclude")
+                .short("e")
+                .long("exclude")
+                .help("Specify package to exclude from sources.")
```
andreaskurth:
```suggestion
                .help("Specify package to exclude from sources")
```

src/cmd/sources.rs
```diff
@@ -33,20 +34,55 @@ pub fn new<'a, 'b>() -> App<'a, 'b> {
                 .long("flatten")
                 .help("Flatten JSON struct"),
         )
+        .arg(
+            Arg::with_name("package")
+                .short("p")
+                .long("package")
+                .help("Specify package to show sources for.")
```
andreaskurth:
```suggestion
                .help("Specify package to show sources for")
```

src/cmd/sources.rs
```diff
@@ -33,20 +34,55 @@ pub fn new<'a, 'b>() -> App<'a, 'b> {
...
+                .takes_value(true)
+                .multiple(true)
+                .number_of_values(1),
+        )
+        .arg(
+            Arg::with_name("no_deps")
+                .short("n")
+                .long("no_deps")
+                .help("Exclude all dependencies, i.e. only top level or specified package(s)."),
```
andreaskurth:
```suggestion
                .help("Exclude all dependencies, i.e. only top level or specified package(s)"),
```

src/cmd/sources.rs
```diff
@@ -33,20 +34,55 @@ pub fn new<'a, 'b>() -> App<'a, 'b> {
...
+        )
+        .arg(
+            Arg::with_name("exclude")
+                .short("e")
+                .long("exclude")
+                .help("Specify package to exclude from sources.")
```
andreaskurth:
```suggestion
                .help("Specify package to exclude from sources")
```

src/src.rs
```diff
@@ -101,6 +103,108 @@ impl<'ctx> SourceGroup<'ctx> {
...
+                    .iter()
+                    .filter_map(|file| match *file {
+                        SourceFile::Group(ref group) => group
+                            .filter_packages(packages)
+                            .map(|g| SourceFile::Group(Box::new(g))),
+                        ref other => Some(other.clone()),
+                    })
+                    .collect();
+            }
+        }
```
andreaskurth: A `match` statement here leads to redundant code. I suggest to instead use a classic `if` with short-circuit evaluation:
```suggestion
        if self.package.is_none() || packages.contains(self.package.unwrap()) {
            files = self
                .files
                .iter()
                .filter_map(|file| match *file {
                    SourceFile::Group(ref group) => group
                        .filter_packages(packages)
                        .map(|g| SourceFile::Group(Box::new(g))),
                    ref other => Some(other.clone()),
                })
                .collect();
        }
```

accuminium (changes requested):
Useful extension that overall LGTM, thanks! I have some suggestions for changes

accuminium:
While experimenting with this extension, I found that include directories are currently not affected by the filter mechanisms.

For instance, in the `axi` repo
```sh
$ bender script vsim -e common_cells
# ...
if {[catch {vlog -incr -sv \
    +define+TARGET_SIMULATION \
    +define+TARGET_VSIM \
    "+incdir+$ROOT/.bender/git/checkouts/common_cells-bdd43b87bfd50406/include" \
    "+incdir+$ROOT/include" \
# ...
```
To avoid surprises, we should probably also process the exported include paths with the filtering mechanism. What do you think?

micprog:
> While experimenting with this extension, I found that exported include directories are currently not affected by the filter mechanisms.
> 
> For instance, in the `axi` repo
> 
> ```shell
> $ bender script vsim -e common_cells
> # ...
> if {[catch {vlog -incr -sv \
>     +define+TARGET_SIMULATION \
>     +define+TARGET_VSIM \
>     "+incdir+$ROOT/.bender/git/checkouts/common_cells-bdd43b87bfd50406/include" \
>     "+incdir+$ROOT/include" \
> # ...
> ```
> 
> To avoid surprises, we should probably also process the exported include directories with the filtering mechanism. What do you think?

My original approach to the include directories was not to worry about them, as they do not specifically reference files, but may have important information for other repositories. As a specific example, the pulp repository exports the `rtl/includes` directory, for which files are included in dependencies such as `pulp_soc` and `pulp_cluster`. Therefore, if you would e.g. only like to synthesize the `pulp_cluster` (`$ bender script synopsys -p pulp_cluster`) and only source these components, removing the include directories form other dependencies would mean not all required include directories are present. 

However, as this is a rather niche use-case with a reasonable workaround (add the include directory to the script manually), and a counter example is also possible (dependency and includes are sourced separately), I agree with the suggestion to keep the behavior consistent. I will make the change, do you think adding a flag (e.g. `--all-include-dirs`) to keep the previous behavior would make sense?

accuminium:
> My original approach to the include directories was not to worry about them, as they do not specifically reference files, but may have important information for other repositories. As a specific example, the pulp repository exports the `rtl/includes` directory, for which files are included in dependencies such as `pulp_soc` and `pulp_cluster`. Therefore, if you would e.g. only like to synthesize the `pulp_cluster` (`$ bender script synopsys -p pulp_cluster`) and only source these components, removing the include directories form other dependencies would mean not all required include directories are present.

I consider *exporting a directory for use by dependencies* an anti-pattern, because it causes a dependency to effectively depend on its parent. (Configuration should be passed top-down through module parameters. Note that [the coding guidelines we use](https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#parameterized-objects-modules-etc) state "`define` and `defparam` should never be used to parametrize a module".) I think this anti-pattern should not guide our design decisions here.

> However, as this is a rather niche use-case with a reasonable workaround (add the include directory to the script manually), and a counter example is also possible (dependency and includes are sourced separately), I agree with the suggestion to keep the behavior consistent.

Indeed, if users really want to go down that road, they can manually work around the default behavior.

> I will make the change, do you think adding a flag (e.g. `--all-include-dirs`) to keep the previous behavior would make sense?

No, I simply do not see any sane use case that would require this.

micprog:
Changing the behavior for include directories was a bit more involved than expected, as it was no longer as easy as filtering out a group. This branch now has a related change, limiting the `export_include_dir` to adding incdirs to its own files, as well as packages directly citing this dependency, no longer adding this globally (this handles an open TODO in the code). This may break projects not properly citing the required dependencies.

accuminium (approved):
LGTM, thanks for implementing the discussed changes!

PR #72, opened 2022-01-18
Fix panic when calling parents

micprog:
In case a dependency doesn't have a manifest, the parents command panicks as it cannot read the dependencies from the manifest. This fixes the issue, simply ignoring these dependencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 ### Changed
 - Interpret all dependency names as lowercase to reduce ambiguity
 
+### Fixed
+- Fix panic of `parents` command if a dependency does not have a manifest
+
 ## 0.24.0 - 2022-01-06
 ### Added
 - Add error if a cyclical dependency is detected to avoid infinite loop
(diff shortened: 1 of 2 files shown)


accuminium (approved):
LGTM, thanks!

accuminium (commented):
Would it not be cleaner to return an early `Err` from the `run` function of the `parents` subcommand, instead of inserting dummy entries as in this implementation?

micprog:
> Would it not be cleaner to return an early `Err` from the `run` function of the `parents` subcommand, instead of inserting dummy entries as in this implementation?

I agree the dummy entries are not the cleanest solution, however I would refrain from explicitly returning errors here as it would reduce the usability of the `parents` command. The problem mainly arises when the `Bender.yml` files are out of sync with a repository making use of the `checkout_dir` functionality. In this case, it is possible that a bender update will add certain dependencies to the lock file that are no longer required by the checked out dependencies, resulting in a missing link. I think the most sensible solution would be to properly update the resolution if `checkout_dir` is used, ensuring these `Bender.yml` files are used for resolution instead of the upstream git dependencies. Once this is fixed, the missing links should disappear and no longer result in dummy entries, so these missing links can return errors as they should no longer be called. For now, however, I would like to keep a functionality like this (could be altered to warnings), such that the feature is still usable and does not by default result in an error, removing even the partial functionality.

accuminium:
I understand and agree, thanks!

micprog:
Added an issue #74 to track this problem, thanks for the review!

PR #73, opened 2022-01-18
Dependency names are lowercase to reduce ambiguity

micprog:
Interpret all dependency names as lowercase to allow for custom capitalization, but no difference of the package (i.e. `fpnew` == `FPnew` or `axi` == `AXI`). May also increase compatibility for non-case-sensitive file systems.


accuminium (approved):
LGTM, thanks!

PR #79, opened 2022-03-21
Git permissions

micprog:
Add various warnings to provide hints for why a git dependency may be failing. Usually this error will relate to a bad git url, bad access permissions, or a missing ssh key on the git server. Other reasons are possible, but this should provide some hints to a new user.


andreaskurth (approved):
LGTM, thanks!

PR #80, opened 2022-03-25
Update parents command to show source mismatch and warn about override

micprog:


andreaskurth (commented):
Could you please add an entry to the Changelog that describes this change? I find it difficult to review this PR and to retrace whether the implementation is doing what it is supposed to do when I don't know what it is supposed to do ;-)

micprog:
Oversight on my part, sorry about that, I added a changelog entry and will squash when merging.

andreaskurth (commented):
LGTM, I only have one suggestion to homogenize and deduplicate code

src/sess.rs
```diff
@@ -1370,6 +1370,17 @@ pub enum DependencySource {
     Git(String),
 }
 
+impl<'a> From<&'a config::Dependency> for DependencySource {
+    fn from(cfg: &'a config::Dependency) -> DependencySource {
+        match *cfg {
+            config::Dependency::Path(ref path) => DependencySource::Path(path.clone()),
+            config::Dependency::GitRevision(ref url, _) => DependencySource::Git(url.clone()),
+            config::Dependency::GitVersion(ref url, _) => DependencySource::Git(url.clone()),
+            config::Dependency::Version(_) => DependencySource::Git("not linked".to_string()),
```
andreaskurth, line 1379: I think this should be
```suggestion
            config::Dependency::Version(_) => DependencySource::Registry,
```
like here:
https://github.com/pulp-platform/bender/blob/22784177e742d1bfb1543969a1241f6d1bd6587c/src/sess.rs#L133-L136

I suggest to update the `impl From` with the latter code and use the `From` trait where the latter code is currently used.

andreaskurth: I think this should be
```suggestion
            config::Dependency::Version(_) => DependencySource::Registry,
```
like here:
https://github.com/pulp-platform/bender/blob/22784177e742d1bfb1543969a1241f6d1bd6587c/src/sess.rs#L133-L136

I suggest to update the `impl From` with the latter code and use the `From` trait where the latter code is currently used.

micprog: Makes sense, I have updated the corresponding sections.

andreaskurth: Thanks!

andreaskurth (approved):

PR #75, opened 2022-01-28
Properly source dependency manifests

micprog:
This PR properly links dependencies in a checkout_dir if present, emulating override path dependencies. This allows for a consistent dependency tree after `bender update` is run, as the local manifests are used to assign dependencies for each package, instead of a possibly out of date upstream version.

PR #81, opened 2022-04-01
Fix panic with checkout_dir if dir does not exist

micprog:
When running `bender update`, the command may currently panic if the `checkout_dir` directory optionally specified in the top-level `Bender.yml` file does not yet exist

### DIFF
--- a/src/resolver.rs
+++ b/src/resolver.rs
@@ -69,18 +69,20 @@ impl<'ctx> DependencyResolver<'ctx> {
         // Store path dependencies already in checkout_dir
         match self.sess.manifest.workspace.checkout_dir.clone() {
             Some(checkout) => {
-                for dir in fs::read_dir(checkout).unwrap() {
-                    self.checked_out.insert(
-                        dir.as_ref()
-                            .unwrap()
-                            .path()
-                            .file_name()
-                            .unwrap()
-                            .to_str()
-                            .unwrap()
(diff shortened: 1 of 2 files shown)


micprog:
The problematic lines were introduced in #75 and should not be an issue yet in any releases.

src/resolver.rs
```diff
@@ -69,18 +69,20 @@ impl<'ctx> DependencyResolver<'ctx> {
...
-                            .unwrap()
-                            .path()
-                            .file_name()
-                            .unwrap()
-                            .to_str()
-                            .unwrap()
-                            .to_string(),
-                        config::Dependency::Path(dir.unwrap().path()),
-                    );
+                if checkout.exists() {
```
andreaskurth, line 72: If this is not true, no `checkout_dir` will be used, correct? Is this what we want, or would we want Bender to (try to) create the checkout dir in that case?

micprog: The creation and filling of the `checkout_dir` happens at a later stage, the code here only checks for dependencies already there. This appropriately updates other dependencies based on changes to the manifest within these folders (e.g. adding a new dependency to the `Bender.yml`), but is not needed for the first step of checking out the files.

andreaskurth: I see, thanks for the explanation!

andreaskurth (approved):

PR #82, opened 2022-04-05
Fix incomplete dependency version update when refetching from remote

micprog:
When releasing a new version of a dependency and updating the reference in a `Bender.yml` file, bender will not immediately find the correct reference when calling `bender update`, although it does fetch the dependency. Calling `bender update` again will fix this, but it should be done in the first step, without causing an error.

This improvement was introduced in v0.22.0 with b42ed72b9dac7b27b1da62c8649b09f558c7982f, but the implementation is incomplete. This change should properly update the dependency list.


micprog:
I was just able to properly test this, the feature behaves as expected now with the fix.

PR #84, opened 2022-04-07
Add checkout command

micprog:
Add `checkout` command to fetch and checkout dependencies. This is automatically taken care of by any script generation/sources command, however some users prefer an explicit command.


src/cmd/checkout.rs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2021 ETH Zurich
+// Michael Rogenmoser <michaero@student.ethz.ch>
```
andreaskurth: Nit: I suggest you use your non-student address.

andreaskurth (approved):
LGTM, thanks! I only have one nit.

Issue #85, opened 2022-04-08
Version mismatch selection fails


micprog:
Likely due to #83, the prompt from bender to select a version when there are conflicting requirements may not use the selected version.

micprog:
Fixed in #86

PR #86, opened 2022-04-08
Enforce version selection consistency (#85)

micprog:
Save the DependencyConstraint instead of an ID to ensure the proper version remains consistent when selecting one manually.

This issue is quite difficult to reproduce, and may remain hidden as the behavior is not very consistent.


andreaskurth (approved):
LGTM, thanks!

PR #87, opened 2022-04-11
Update `Bender.lock` when running `bender clone`

### LINKED ISSUE #46: bender clone command should implicitly update dependency path
Trying out the new bender `clone` command I noticed that while bender clone <package> name successfully checks out the package and registers it in the Bender.local file it does not update the path corresponding to the just cloned package. This can be forced with a `bender update` but intuitively I would have expected that:

`bender clone my_package`
immediately followed by
`bender path my_package`

would yield `working_dir/my_package`. However, right now the path command still returns the hidden internal git directory until I manually call `bender update`...

micprog:
This removes the requirement to run `bender update` after cloning a dependency. Fixes #46.

### DIFF
--- a/src/cmd/clone.rs
+++ b/src/cmd/clone.rs
@@ -1,5 +1,5 @@
 // Copyright (c) 2021 ETH Zurich
-// Michael Rogenmoser <michaero@student.ethz.ch>
+// Michael Rogenmoser <michaero@iis.ee.ethz.ch>
 
 //! The `clone` subcommand.
 
@@ -10,6 +10,7 @@ use std::process::Command;
 use tokio_core::reactor::Core;
 
 use crate::config;
+use crate::config::{Locked, LockedSource};
 use crate::error::*;
 use crate::sess::{Session, SessionIo};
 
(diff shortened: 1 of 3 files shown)


src/cmd/clone.rs
```diff
@@ -211,5 +212,24 @@ pub fn run(sess: &Session, path: &Path, matches: &ArgMatches) -> Result<()> {
 
     println!("{} dependency added to Bender.local", dep);
 
+    use std::fs::File;
+    let file = File::open(path.join("Bender.lock"))
+        .map_err(|cause| Error::chain(format!("Cannot open lockfile {:?}.", path), cause))?;
```
andreaskurth, line 217: `File::open` errors if `Bender.lock` does not already exist. Is it guaranteed to always exist at this point? If not, would we not want to create it at this point rather than failing?

andreaskurth (commented):
LGTM, thanks! I have one question, though

src/cmd/clone.rs
micprog: I'm fairly certain the `Bender.lock` should exist here, as the command will only execute once the main cli function has updated its dependencies: https://github.com/pulp-platform/bender/blob/2ed4671357e3535d7641c88a58bd8c78229b438a/src/cli.rs#L133-L141
https://github.com/pulp-platform/bender/blob/2ed4671357e3535d7641c88a58bd8c78229b438a/src/cli.rs#L227

andreaskurth: I see, thanks for the explanation!

andreaskurth (approved):

PR #88, opened 2022-04-13
Update dependencies

micprog:
Update dependencies, with slight restructuring updating `clap` to v3.1, bumps minimum rust to 1.54


src/cli.rs
```diff
@@ -20,34 +20,32 @@ use tokio_core::reactor::Core;
 
 /// Inner main function which can return an error.
 pub fn main() -> Result<()> {
-    let app = App::new(env!("CARGO_PKG_NAME"))
-        .setting(AppSettings::AllowExternalSubcommands)
+    let app = Command::new(env!("CARGO_PKG_NAME"))
+        // .setting(Command::is_allow_external_subcommands_set)
```
andreaskurth: From [`clap`'s doc](https://docs.rs/clap/latest/clap/struct.App.html#method.allow_external_subcommands) I assume we want `.allow_external_subcommands(true)` here. Or do we no longer need this setting?

Cargo.toml
```diff
@@ -20,14 +20,14 @@ tokio-core = "0.1.12"
...
+clap = { version = "3.1", features = ["derive"] }
+semver = "1.0"
+blake2 = "0.10"
 typed-arena = "2"
-dirs = "3"
+dirs = "4"
 pathdiff = "0.2"
 common-path = "1.0.0"
-itertools = "0.9"
+itertools = "0.10"
```
andreaskurth, line 30: [According to deps.rs](https://deps.rs/repo/github/pulp-platform/bender), the dependencies `futures` and `tokio-process` are also outdated. Is it possible to update them as well, or is this out of scope and should be tracked in a new issue?

src/cmd/script.rs
```diff
@@ -15,20 +15,20 @@ use common_path::common_path_all;
...
         .arg(
-            Arg::with_name("target")
-                .short("t")
+            Arg::new("target")
+                .short('t')
                 .long("target")
                 .help("Only include sources that match the given target")
                 .takes_value(true)
-                .multiple(true)
+                .multiple_values(true)
```
andreaskurth: Should `.multiple_occurrences(true)` also be added here, like for the `path` command? [According to the `clap` doc](https://docs.rs/clap/2.33.3/clap/struct.Arg.html#method.multiple):
> Setting `multiple(true)` for an [option](https://docs.rs/clap/2.33.3/clap/struct.Arg.html#method.takes_value) with no other details, allows multiple values **and** multiple occurrences [...]

The same question raises below for the arguments `define`, `vcom-arg`, and `vlog-arg`, where we had `multiple(true)`, and for `vlogan-bin` and `vhdlan-bin`, where we had `multiple(false)`.

src/cmd/script.rs
```diff
@@ -45,109 +45,108 @@ pub fn new<'a, 'b>() -> App<'a, 'b> {
...
                 .long("no-abort-on-error")
                 .help("Do not abort analysis/compilation on first caught error (only for programs that support early aborting)")
         )
         .arg(
-            Arg::with_name("compilation_mode")
+            Arg::new("compilation_mode")
                 .long("compilation-mode")
                 .help("Choose compilation mode for Riviera-PRO option: separate/common (Riviera-PRO only)")
                 .takes_value(true)
-                .multiple(false)
```
andreaskurth, line 122: Did you remove this because it is the default?

src/cmd/script.rs
```diff
@@ -45,109 +45,108 @@ pub fn new<'a, 'b>() -> App<'a, 'b> {
...
         .arg(
-            Arg::with_name("package")
-                .short("p")
+            Arg::new("package")
+                .short('p')
                 .long("package")
                 .help("Specify package to show sources for")
                 .takes_value(true)
-                .multiple(true)
+                .multiple_occurrences(true)
```
andreaskurth: What about `multiple_values(true)`?

Also below for the `exclude` argument.

src/cmd/sources.rs
```diff
@@ -16,46 +16,46 @@ use crate::src::SourceGroup;
...
         .arg(
-            Arg::with_name("target")
-                .short("t")
+            Arg::new("target")
+                .short('t')
                 .long("target")
                 .help("Filter sources by target")
                 .takes_value(true)
-                .multiple(true)
+                .multiple_occurrences(true)
```
andreaskurth: What about `multiple_values(true)`?

Also below for the `package` and the `exclude` argument.

andreaskurth (commented):
Thanks for this work! I have a questions / discussion points.

src/cli.rs
micprog: Yes, I was unsure about this and forgot to get back to it.

Cargo.toml
micprog: I started looking into this but the breaking changes in [`tokio`](https://docs.rs/tokio/latest/tokio/) and `futures` are significant and would require rewriting large parts of bender, and I don't currently have the time to dig into this.

andreaskurth: Okay, makes sense. Shall we open an issue to track this?

src/cmd/script.rs
micprog: I have updated the change to the `multiple` option. As far as I can tell, the following should hold:
- `.multiple(true).number_of_values(1)` -> `.multiple_occurrences(true)`
- `.multiple(true)` -> `.multiple_values(true).multiple_occurrences(true)`
For some things, `multiple_occurrences` does not make much sense, and if only a single value is needed any reference to the multiple can be removed, simplifying the code. I hope these changes now make more sense, thanks for catching them!

src/cmd/script.rs
micprog: yes

src/cmd/script.rs
micprog: I would suggest to keep this as is, forcing users to explicitly indicate individual packages with `-p package1 -p package2` (similar to the now updated target). Same for below.

Cargo.toml
micprog: Done: #89

src/cmd/script.rs
andreaskurth: Okay, but this is a breaking change, right? If so, please mention it in the Changelog.

andreaskurth: Ah no, we had `number_of_values(1)` before. Okay, then it does not change for the user.

src/cmd/script.rs
andreaskurth: Yes, that makes sense. Thanks for fixing!

andreaskurth (approved):

PR #93, opened 2022-07-08
Add formality support

zarubaf:
The formality script is slightly so different to the `dc_shell` syntax. This also removes the `define` option which has no meaning for VHDL files (and becomes an error in `formality`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+
+### Added
+- Added `formality` script.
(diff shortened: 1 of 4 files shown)


andreaskurth (approved):
LGTM, thanks @zarubaf!

micprog:
Thanks @zarubaf! I'm not very familiar with formality, but would it be necessary to have the bender script switchable between the `-r` reference mode and the `-i` implementation mode? If not I'll merge.

zarubaf:
I am not terribly familiar myself tbh, but my understanding was that the RTL will always go into the reference bucket. Netlists or other output artifacts that you want to compare you likely do not want to pull-in via `bender` (IMHO at least).

micprog:
Ok, if needed a later update can integrate this feature as well. Merging

zarubaf:
@micprog Do you think you could release a new version of `bender` with the formality support? The pre-built artifacts are very useful for air-gapped machines :-)

PR #94, opened 2022-07-28
Update tokio, update futures, rewrite for async/await

micprog:
To update further dependencies, the codebase was slightly modified to comply with `async/await`, updating tokio and futures dependencies. Most of the structures were attempted to be kept, but some were modified away from `.and_then` structures for simpler implementation. Due to time constraints, throttling was not fully implemented yet. My own testing did not uncover any issues, also on larger (PULPissimo) repositories, but improved throttling implementation may be necessary.

### DIFF
--- a/src/sess.rs
+++ b/src/sess.rs
@@ -16,18 +16,15 @@ use std::sync::atomic::AtomicUsize;
 use std::sync::{Arc, Mutex};
 use std::time::SystemTime;
 
+use crate::futures::{FutureExt, TryFutureExt};
 use futures::future::{self, join_all};
-use futures::Future;
-use semver;
-use serde_yaml;
-use tokio_core::reactor::Handle;
 use typed_arena::Arena;
 
 use crate::cli::read_manifest;
 use crate::config::Validate;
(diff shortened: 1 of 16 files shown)


src/sess.rs
```diff
@@ -495,103 +490,86 @@ impl<'io, 'sess: 'io, 'ctx: 'sess> SessionIo<'sess, 'ctx> {
...
-                        warnln!("Please ensure the url is correct and you have access to the repository.");
-                        Error::chain(
-                            format!("Failed to initialize git database in {:?}.", db_dir),
-                            cause,
-                        )
-                    })
-                    .map(move |_| git),
-                ),
-            )
+            // TODO MICHAERO: May need throttle
```
andreaskurth, line 500: Suggest to create an issue on GitHub to track this and the other throttle

andreaskurth (approved):
Thanks for this big piece of work, @micprog!

I could not review this in detail due to time constraints, but if you have tested the changes with sufficiently complex use cases, I suggest to proceed with merging.

PR #98, opened 2022-08-17
Fix path subdependencies in git dependencies

micprog:
- Partially checkout path dependencies in git dependencies
- Add warnings if manifest file is not found for a dependency


paulsc96 (approved):
LGTM :+1:

PR #99, opened 2022-08-22
Add `import` command to import files from non-bender repositories

micprog:
This command can import source files from other git repositories, similar to lowrisc's `vendor.py` script. Check the README for more details. An example use case can be found in the `bender_vendor` branch in [`register_interface`](https://github.com/pulp-platform/register_interface/tree/bender_vendor)


paulsc96 (approved):
LTGM, works :+1:

Not sure on the clippy unwrap warning, but if it is intentional, we may want to disable it.

PR #100, opened 2022-08-26
Fix: Keep export_incdirs even when excluding deps

micprog:
Include directories may be needed to properly parse files, so this PR enforces keeping `export_incdirs` for dependencies when excluding dependencies, e.g. with `-n` or `-e` flags.


thommythomaso (approved):
LGTM, Works as expected with the LLC!

PR #102, opened 2022-09-13
Update script for rhel and missing lsb_release

micprog:
As [rhel9 (and thus also almalinux 9) no longer supports `lsb_release`](https://bugzilla.redhat.com/show_bug.cgi?id=2012924), this is a workaround for these distributions (and others that may not support `lsb_release` in the future), fetching the needed information from `/etc/os-release`.

### DIFF
--- a/init
+++ b/init
@@ -36,7 +36,7 @@ main() {
     fi
 
     downloader --check
-    need_cmd lsb_release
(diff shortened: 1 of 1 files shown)


bluewww (approved):
LGTM

PR #105, opened 2022-10-31
import: Streamline init of external dependency

niwis:
Streamline the initialization of an external dependency. In particular:

- Catch missing target dir
- Create initial patch if none exist


micprog (approved):
LGTM, thanks for the fix! 👍

PR #106, opened 2022-11-04
init: Use bash

bluewww:
This script is not fully POSIX-compliant (see shellcheck complaints). Easiest fix is to just use bash.


micprog (approved):
LGTM :+1:, thanks!

PR #108, opened 2022-12-14
import: Restructure and extend functionality

niwis:
Refactor and extend the `import` subcommand. The most significant changes:
- Rename the subcommand to `vendor`.
- Use sub-sub commands (`init`, `diff`, `patch`) for main functionality. One sub-sub command has to be provided (otherwise the help message will appear)
- Emit format-patches per default for more control over patch contents and improved trackability and upstreamability. The old behaviour (plain diff) can be achieved using the `--plain` flag.
- Make mappings optional (actually they should be used with caution). If no mappings are provided, the directory structure of the dependency will be kept and copied to `target_dir`.
- Add an optional `include_from_upstream` field to specify patterns that need to match the items of the dependency that are to be copied.

More information on the usage in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -483,7 +483,7 @@ If the `--plain` argument is passed, this command will *not* prompt for a commit
 #### Example workflow
 
 Let's assume we would like to vendor a dependency `my_ip` into a project `monorepo`.
-A simple configuration ind `Bender.yml` could look as follows (see the `Bender.yml` description above for more information on this):
+A simple configuration in a `Bender.yml` could look as follows (see the `Bender.yml` description above for more information on this):
 
 ```yaml
 vendor_package:
@@ -501,7 +501,7 @@ We can print these changes with the command `bender vendor diff`.
 Now, we would like to generate a patch with the changes in `deps/my_ip/a` (but not those in `deps/my_ip/b`).
 We stage the desired changes using `git add deps/my_ip/a` (of course, you can also just stage parts of a file using `git add --patch`).
 The command `bender vendor patch` will now ask for a commit message that will be associated with this patch.
-Then, it will place a patch that contains our changes in `deps/my_ip/a` into `deps/patches/my_ip/0001-commit-message.patch`
+Then, it will place a patch that contains our changes in `deps/my_ip/a` into `deps/patches/my_ip/0001-commit-message.patch` (the number will increment if a numbered patch is already present).
 
 We can easily create a corresponding commit in the monorepo.
 `deps/my_ip/a` is still staged from the previous step.
--- a/README.md
+++ b/README.md
@@ -176,21 +176,24 @@ workspace:
 plugins:
   hello: scripts/hello.sh
 
-# List of imported files from external repositories not supporting bender. Optional
-external_import:
+# List of vendorized files from external repositories not supporting bender. Optional.
+vendor_package:
     # package name
   - name: lowrisc_opentitan
     # target directory
     target_dir: vendor/lowrisc_opentitan
     # upstream dependency (i.e. git repository similar to dependencies)
     upstream: { git: "https://github.com/lowRISC/opentitan.git", rev: "47a0f4798febd9e53dd131ef8c8c2b0255d8c139" }
-    # paths to exclude from upstream dependency
+    # paths to include from upstream dependency. Per default, all paths are included. Optional.
+    include_from_upstream:
+      - "src/*"
+    # paths to exclude from upstream dependency. Paths that also match a pattern in include_from_upstream are excluded. Optional.
     exclude_from_upstream:
       - "ci/*"
-    # directory containing patch files
+    # directory containing patch files. Optional.
     patch_dir: "vendor/patches"
-    # file mapping from remote repository to local repository, with optional patch_dir containing patches
-    mapping: 
+    # custom file mapping from remote repository to local repository, with optional patch_dir containing patches. Optional. Note: mappings make upstreaming patches slightly more complicated. Avoid if not necessary.
+    mapping:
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -454,16 +457,64 @@ The `bender parents <PKG>` command lists all packages calling the `PKG` package.
...
+#### `vendor patch` --- Generate a patch file from local changes
+
+If there are local, *staged* changes in a vendored dependency, this command prompts for a commit message and generates a patch for that dependency. The patch is written into `patch_dir`.
+
+If the `--plain` argument is passed, this command will *not* prompt for a commit message and generate a patch of *all* (staged and unstaged) local changes of the vendored dependency.
+
+#### Example workflow
+
+Let's assume we would like to vendor a dependency `my_ip` into a project `monorepo`.
+A simple configuration ind `Bender.yml` could look as follows (see the `Bender.yml` description above for more information on this):
```
micprog:
```suggestion
A simple configuration in a `Bender.yml` could look as follows (see the `Bender.yml` description above for more information on this):
```

README.md
```diff
@@ -454,16 +457,64 @@ The `bender parents <PKG>` command lists all packages calling the `PKG` package.
...
+
+Executing `bender vendor init` will now clone this dependency from `upstream` and place it in `target_dir`.
+
+Next, let's assume that we edit two files within the dependency, `deps/my_ip/a` and `deps/my_ip/b`.
+We can print these changes with the command `bender vendor diff`.
+
+Now, we would like to generate a patch with the changes in `deps/my_ip/a` (but not those in `deps/my_ip/b`).
+We stage the desired changes using `git add deps/my_ip/a` (of course, you can also just stage parts of a file using `git add --patch`).
+The command `bender vendor patch` will now ask for a commit message that will be associated with this patch.
+Then, it will place a patch that contains our changes in `deps/my_ip/a` into `deps/patches/my_ip/0001-commit-message.patch`
```
micprog: Are the numbers of the patch incrementing? Should this be mentioned here for clarity?

meggiman:
I tried this branch out a bit. There are some points that could be improved:
- If possible, using sparse cloning could decrease runtime significantly, especially for large mono-repositories.
- The existing patching logic assumes all mappings to be from a directory to a directory. However, some IPs that already use the vendor feature (pulp-platform/register_interface) use a mapping for only a single file. The new flow no longer applies to these use cases since the `mapping.from` key is directly supplied to the `--directory` flag of `git am/git apply` without checking if it is really a directory. We should discuss how we handle these cases.

src/cmd/vendor.rs
```diff
@@ -0,0 +1,590 @@
...
+        .subcommand(Command::new("init")
+            .about("(Re-)initialize the external dependencies. Copies the upstream files into the target directories and applies existing patches.")
+            .arg(
+                Arg::new("no_patch")
+                    .short('n')
+                    .long("no_patch")
+                    .help("Do not apply patches when initializing dependencies"),
+            )
+        )
+        .subcommand(Command::new("patch")
```
micprog, line 59: Would it make sense to also add an optional argument `-m <msg>` to directly pass the message?

micprog (commented):
Thank you very much for the contribution! Overall it looks quite good, there are some nits and some minor issues that still need to be investigated.

- [ ] The current behavior when vendoring/copying the files puts them into the current directory, not the top directory specified by bender (e.g. when selecting a different directory for bender with `-d`).

micprog:
> * The existing patching logic assumes all mappings to be from a directory to a directory. However, some IPs that already use the vendor feature (pulp-platform/register_interface) use a mapping for only a single file. The new flow no longer applies to these use cases since the `mapping.from` key is directly supplied to the `--directory` flag of `git am/git apply` without checking if it is really a directory. We should discuss how we handle these cases.

I attempted to fix at least the application of patches in the newest commit, the patch should now be able to tell whether or not a file or directory is passed, linking the parent directory for files.

bluewww:
Currently, I don't have the bandwidth to review this feature but I would definitely have a use case for it

src/cmd/vendor.rs
```diff
@@ -0,0 +1,645 @@
...
+                    }
+                    Ok(())
+                })
+            }
+
+            Some(("init", matches)) => {
+                patch_links.clone().into_iter().try_for_each(|patch_link| {
+                    stageln!("Copying", "{} files from upstream", vendor_package.name);
+                    // Remove existing directories before importing them again
+                    std::fs::remove_dir_all(
```
meggiman: wrap this with std::fs::try_exists and if it doesn't, create with std::fs::create_dir_all. Also we should explicitly handle the None case. Just using unwrap is kind of an anti-pattern I heard and the corresponding panik message is often not very useful for the user to figure out what was going wrong.

src/cmd/vendor.rs
```diff
@@ -0,0 +1,645 @@
...
+    // import necessary files from upstream, apply patches
+    let dep_path = dep_path.as_ref();
+
+    // Make sure the target directory actually exists
+    let link_to = patch_link
+        .to_prefix
+        .clone()
+        .prefix_paths(&vendor_package.target_dir);
+    let link_from = patch_link.from_prefix.clone().prefix_paths(dep_path);
+    std::fs::create_dir_all(&link_to.parent().unwrap())?;
```
meggiman: Again, we should print some nicer error message if directory creation fails (e.g. due to issues with write permissions).

meggiman (commented):
I am not familiar enough with Rust to give an in-depth code review. Just nitpicking about a panik I got because target dirs didn't exist yet.

meggiman (approved):
Now it seems to work for all of my use cases. Thanks a lot!

micprog (approved):
LGTM :+1: , merging

PR #109, opened 2022-12-22
Lockfile: use relative paths if possible

micprog:
Use relative paths in Lockfile if path dependency is in a subdirectory


SamuelRiedel (approved):
LGTM :+1:

Issue #116, opened 2023-01-27
Rerunning the same bender command twice produces different results


epsilon537:
When I run 'bender script verilator' twice on my source tree, I get slightly different results. The files are reordered:

First time:
```
...
+define+TARGET_SYNTHESIS
+define+TARGET_VERILATOR
+incdir+/home/epsilon/work/boxlambda/gw/components/ibex/../../../sub/ibex/build/lowrisc_ibex_top_artya7_0.1/src/lowrisc_
prim_util_get_scramble_params_0/rtl
+incdir+/home/epsilon/work/boxlambda/gw/components/ibex/../../../sub/ibex/build/lowrisc_ibex_top_artya7_0.1/src/lowrisc_
ibex_ibex_core_0.1/rtl
+incdir+/home/epsilon/work/boxlambda/gw/components/ibex/../../../sub/ibex/vendor/lowrisc_ip/ip/prim/rtl
+incdir+/home/epsilon/work/boxlambda/gw/components/ibex/../../../sub/ibex/vendor/lowrisc_ip/dv/sv/dv_utils
+incdir+/home/epsilon/work/boxlambda/gw/components/ibex/../../../sub/ibex/build/lowrisc_ibex_top_artya7_0.1/src/lowrisc_
prim_util_memload_0/rtl
...

```
Second time:

```
+define+TARGET_SYNTHESIS
+define+TARGET_VERILATOR
+incdir+/home/epsilon/work/boxlambda/gw/components/ibex/../../../sub/ibex/vendor/lowrisc_ip/dv/sv/dv_utils
+incdir+/home/epsilon/work/boxlambda/gw/components/ibex/../../../sub/ibex/build/lowrisc_ibex_top_artya7_0.1/src/lowrisc_
dv_dv_fcov_macros_0
+incdir+/home/epsilon/work/boxlambda/gw/components/ibex/../../../sub/ibex/build/lowrisc_ibex_top_artya7_0.1/src/lowrisc_
prim_util_get_scramble_params_0/rtl
+incdir+/home/epsilon/work/boxlambda/gw/components/ibex/../../../sub/ibex/build/lowrisc_ibex_top_artya7_0.1/src/lowrisc_
prim_util_memload_0/rtl
+incdir+/home/epsilon/work/boxlambda/gw/components/ibex/../../../sub/ibex/build/lowrisc_ibex_top_artya7_0.1/src/lowrisc_
ibex_ibex_core_0.1/rtl
[...]

PR #117, opened 2023-01-28
Use `IndexMap` and `IndexSet` instead of the `std Hash*` equivalents to preserve ordering

micprog:


micprog:
See #116

src/config.rs
```diff
@@ -160,7 +161,7 @@ pub enum SourceFile {
     /// A file.
     File(PathBuf),
     /// A subgroup.
-    Group(Box<Sources>),
```
joennlae: Why exactly do we remove the `Box`? This is again more an understanding question than something that actually matters. I assume we could have a veeeery slight performance lose because of that. Am I right in that assumption?

My ref: 
https://doc.rust-lang.org/book/ch15-01-box.html

I check it would still work with using `Box`

joennlae (approved):
LGTM 💯 

I just have a question regarding the removal of `Box` e.g. allocating sources on the stack instead of the heap. Why are we exactly doing that?

src/config.rs
micprog: I think it makes sense to keep the box, so I reverted the commits. Thanks for the quick review!

PR #119, opened 2023-03-23
Fix overrides dependency name

micprog:
Parse override dependencies in lowercase to align to change in 0.25.0


niwis (approved):
Thanks a lot, LGTM!

PR #120, opened 2023-03-23
Templates for scripting

### LINKED ISSUE #124: Defines and Includes in flist
Bender doesn't
* emit defines
* miss includes in dependencies

when generating flists

micprog:
Modify script output to use custom templates (tera package). Allows use of custom script template files.

Fixes #124


bluewww (approved):
LGTM.

I tested `flist` (it also fixes #124) and `vsim` and `vcs` scripts output in carfield's safety island.

micprog:
Review status for all script types:
- [ ] flist
- [ ] vsim
- [ ] vcs
- [ ] verilator
- [ ] synopsys
- [ ] formality
- [ ] riviera
- [ ] genus
- [ ] vivado
- [ ] vivado-sim
- [ ] precision

jwnrt:
Apologies for what's probably a bit of an obnoxious comment, but I was wondering what made you choose `tera` for the templating?

I ask because the dependency adds 69 extra transitive dependencies and increases the build time by ~50% (measured across 12 cores). Maybe this isn't a metric you're measuring or optimizing for, which is totally fair.

From a quick survey of templating engines, [tinytemplate] looks like the most popular and you would hope is lightweight. I've only skimmed this PR's template usage, so maybe this doesn't have all the features you need, though.

This is a great feature regardless, thanks for all the work you're doing on it! I don't want to dismiss any of this work or suggest you redo anything!

[tinytemplate]: https://crates.io/crates/tinytemplate

micprog:
> Apologies for what feels like a really obnoxious comment, but I was wondering what lead you to choose `tera` for the templating?
> 
> My motivation for asking is that it feels quite heavy with 68 transitive dependencies and adding ~50% to the build time. Maybe this isn't a metric you're measuring or optimizing for, which is totally fair!
> 
> From a quick survey of templating engines, [tinytemplate](https://crates.io/crates/tinytemplate) is quite widely used and hopefully lightweight, but there are others too. I've only skimmed this PR's template usage, so maybe this doesn't have all the features you need, though.
> 
> This is a great feature regardless, thanks for all the work you're doing on it! I don't want to dismiss any of this work or suggest you redo anything!

Hi @jwnrt,

The main reason I opted for `tera` was its similarity to other popular templating libraries, lowering the barrier of entry for those not familiar with Rust. The main motivation with this change is to make bender easier to use without having to figure out the intricacies within bender. It was also a good opportunity to streamline the script generation flow for all the different available tools.

I understand that the build time increases, but this is mainly a constraint on development - for deployment, there are easily available [binary releases](https://github.com/pulp-platform/bender/releases) with a [custom script](https://github.com/pulp-platform/bender#installation) for downloading. We have even recently added a custom [github CI action](https://github.com/pulp-platform/pulp-actions/tree/main/bender-install) to install bender within github actions. Build time of the tool is not a concern.

Regarding other template libraries, I have heard great things about `mako`, but this is limited to Python. In an effort to allow easier extension to other tools without needing to be familiar with Rust, I opted for a more advanced templating library than `tinytemplate` to allow more advanced features, in case they are necessary (I use [replacements](https://github.com/pulp-platform/bender/blob/abf7b94dd528565c68fa3a7997699ff99ba424c1/src/cmd/script.rs#L616) and [conditions on loop state](https://github.com/pulp-platform/bender/blob/abf7b94dd528565c68fa3a7997699ff99ba424c1/src/cmd/script.rs#L647), which I don't think is possible with tinytemplate). Arguably, some things could be handled within the software, not the template, but the aim was to also provide the option for a user input template without needing to modify the Rust code. Also, not all libraries have equal active development.

Either way, thank you for your question, I hope this clarifies it somewhat. Let me know if you have feedback on the generated scripts with the new template!

jwnrt:
Yep, that answers my question, thank you for the very detailed response!

src/cmd/script.rs
```diff
@@ -781,542 +483,643 @@ fn emit_synopsys_tcl(
...
+    {% if relativize_path %}\
+        {% if file is starting_with(root) %}\
+            {{ file | replace(from=root, to='') | trim_start_matches(pat='/') }}\n\
+        {% else %}\
+            {{ file }}\n\
+        {% endif %}\
+    {% else %}\
+        {{ file }}\n\
+    {% endif %}\
+{% endfor %}";
```
Andful: May this not be easier with `include_str`?

micprog: I originally did this with a string internal to the Rust file for clearer formatting of the template, but found a way for semi-clean formatting with comments also outside of the file, included now with `include_str`. This can also provide a better basis for custom templates, in case they are needed.

CyrilKoe:
Tested vivado script, see [Carfield pipeline](https://iis-git.ee.ethz.ch/github-mirror/carfield/-/pipelines/81129).
Visual inspection of vivado-sim ok, only difference with before is that it now repeats certain defines:
```
set_property verilog_define [list \
    CLUSTER_ALIAS \
    CLUSTER_ALIAS \
    FEATURE_ICACHE_STAT \
    FEATURE_ICACHE_STAT \
    GEN_NO_HYPERBUS=1 \
    HIERARCHY_ICACHE_32BIT \
    HIERARCHY_ICACHE_32BIT \
    ICAHE_USE_FF \
    ICAHE_USE_FF \
    PRIVATE_ICACHE \
    PRIVATE_ICACHE \
    PULP_FPGA_EMUL \
    TARGET_BSCANE \
```
(Not the case before)

fischeti:
Synopsys seems to work fine, I didn't see any issues pop during an elaboration step.

CyrilKoe:
Uniquify works for vivado/vivado-sim, now the script is the same as for previous bender version (just with with some re-ordering)

fischeti:
scripts generated for verilator also seem to run through now

fischeti (approved):

PR #122, opened 2023-04-01
Fix order of from/to in vendor's copy error messages

jwnrt:
Great work with bender!

Just a small fix for the order of these from/to arguments in an error message I was hitting.


src/cmd/vendor.rs
```diff
@@ -433,8 +433,8 @@ pub fn diff(
                 Error::chain(
                     format!(
                         "Failed to copy {} to {}.",
-                        link_to.to_str().unwrap(),
                         link_from.to_str().unwrap(),
+                        link_to.to_str().unwrap(),
```
micprog: I don't think this change here is necessary, as for the `diff` function, the changes are being copied from the `link_to` to the `link_from` (as seen on line 432).

jwnrt: Oh yeah you're right, thanks! I'd just assumed this one was the same as above without reading the context, sorry

jwnrt: Force pushed a new commit without this change.

micprog:
LGTM :+1:, thanks for the fix!

PR #128, opened 2023-06-01
Improve error message when no bender file is found.

hydroid7:
After running into this error, I think a hint never hurts.


micprog:
Merged with 4ead26d56ef1df3211a127624708d5374adac8fa

PR #135, opened 2023-09-07
path: Skip checkout if possible

niwis:
Reduce bender's footprint by skipping checkout of a package if the package path already exists and
- `path` is executed (can be overruled by new `--checkout` flag), or
- for locally linked packages if any command except for `update` is executed.

To prevent `bender path` from returning stale paths, a checkout is now performed automatically after executing `update`. This can be prevented by passing the new `--no-checkout` flag to `update`.


micprog (approved):
LGTM :+1:

PR #147, opened 2023-12-29
Allow environment variables in Bender.yml paths

micprog:


joennlae (approved):
Really nice :-) I tried to check if all paths were replaced. You found all of them, or I could not spot another occurrence of a path that should be substituted.

PR #153, opened 2024-01-23
Fix script when specifying multiple `--only-*`

micprog:
Ensure defines/includes/sources are included in script when specifying multiple `--only-*`.


CyrilKoe (approved):
Ok, tested with Carifled's Vivado block design flow

fischeti (approved):
LGTM. The only suggestion I have is that `--only-*` commands should be mutually exclusive to each other. Changing them to `--no-*` would make more sense to me.

micprog:
> LGTM. The only suggestion I have is that `--only-*` commands should be mutually exclusive to each other. Changing them to `--no-*` would make more sense to me.

While I agree, I think for now, it makes more sense to keep the existing behavior prior to the template change. This just ensures the behavior from before #120 continues working.

PR #155, opened 2024-02-20
Add flist-plus script type to available formats

micprog:
Emits flist with +define+ and +incdir+


fischeti (approved):
`flist-plus` works now as expected. Thanks!

PR #158, opened 2024-03-07
cmd: add 'completion' command

schilkp:
This PR adds a `bender completion` command that generates completion scripts for common shells (bash, zsh, fish, ...)  using clap's clap_complete crate. 

Example: https://asciinema.org/a/V3kPpxB7TCHRS1BJlRfYyhdPT

(Ignore the fact that the recording can't quite deal with unicode and screws up the shell prompt/command :)

It exposes all the shells supported by clap_complete:

- bash
- elvish
- fish
- powershell
- zsh

I am not sure about benders compatibility with windows, it may make sense to remove powershell if it is anyway not supported.

Happy to rework this, if it is not quite fitting in its current state:)


README.md
```diff
@@ -535,6 +535,17 @@ This branch can then be rebased and a pull request can be opened from it as usua
...
+
+The `bender completion <SHELL>` command prints a completion script for the given shell.
+
+Supported shells:
+- `bash`
+- `elvish`
+- `fish`
+- `powershell`
+- `zsh`
+
```
micprog: Can you add how to enable the completion for the different shells?

I think simply mentioning the script needs to be placed into a file and then sourced in the shell, preferably directly in the rc file, would suffice. This should avoid confusion by new users.

micprog (approved):
Other than the minor extension of the README, this LGTM. This feature is very cool, thanks a lot for the contribution!

schilkp:
> Can you add how to enable the completion for the different shells?
> 
> I think simply mentioning the script needs to be placed into a file and then sourced in the shell, preferably directly in the rc file, would suffice. This should avoid confusion by new users.

I have added a note that links to the relevant documentation for the most common shells. I don't know enough about different shells, and would worry that any specific advise is not universally correct :)  It really depends not just on which shell you are using, but also how it is configured (things like oh-my-zsh/bash tend to mess with this sort of thing).  Plus it seems like the kind of info that may become out-of-date quickly..

Let me know if you want more - but if so would probably be good if someone with a little more linux/unix/shell background does that.

On this note: I don't know if there is a "best practice" for submitting such completion scripts to any online repositories (such as https://github.com/zsh-users/zsh-completions), or if they would even accept an auto generated script like this. May be worth investigating.

Similarly I don't know if it would be appropriate/expected behavior for the init.sh script or Arch AUR PKGBUILD to include installation of these completion scripts? I would worry about that being a little brittle/causing problems, but really would improved the user experience.

schilkp:
I also noticed that the 'help'/'about' text for some subcommands/args is rather long. For example 'vendor':

longabout

Splitting those into a short 'about' and full 'long_about' [^1] would make the completions much nicer.

Happy to do that (in here or a separate PR) if desired.

[^1]: https://docs.rs/clap/latest/clap/struct.Command.html#method.long_about

schilkp:
Long helps/abouts have been split. I don't know if there is a 'standard'  for  long help texts: Should they manually be split into lines (to say 80 cols) or should they be on one line and rely on terminal wrapping. At the moment I left them on one line, as they were before.

I also added a note the help text of `bender` and `bender vendor` that explains how to get to the help of a subcommand:

```
...
  init        Initialize a Bender package
  help        Print this message or the help of the given subcommand(s)

Options:
  -d, --dir <dir>  Sets a custom root working directory
      --local      Disables fetching of remotes (e.g. for air-gapped computers)
      --debug      Print additional debug information
  -h, --help       Print help
  -V, --version    Print version

Type 'bender <SUBCOMMAND> --help' for more infromation about a bender subcommand.
```

micprog (approved):
LGTM! Can you rebase on master? Then I will merge.

schilkp:
@micprog  done.

PR #160, opened 2024-04-08
Add fusion-compiler script generator

uge:
This PR adds basic support for Synopsys Fusion Compiler based on the older Synopsys target. The main initial difference is the use of absolute paths throughout the generated script as the Synopsys target template prefixed $ROOT to paths incorrectly. Although this could have been fixed in the Synopsys target, we anticipate additional changes needed for Fusion Compiler specific difference even if the underlying PRESTO parser remains the same.


uge:
> We have also been using Fusion Compiler for some projects, and so far the `synopsys` target provided all we need.

I suspect that would be the case since it's the likely exactly the same front end parser. We don't use dc_shell internally so we have no way to regression test changes for it and I assume it is still much more widely used that Fusion Compiler for now. At a very basic level, splitting it means that further contributions wouldn't have the potential to break the more widely used tool.

In a broader level, we're looking at adding some mixture of generation for physical related info that's more commonly used in a RTL->GDS flow. Bender is neat for what it does, but until the world becomes more rust-like, it's also questionable if we'll have internal resources to implement the changes in bender or will do it outside even though the natural place would be in the dependency management system. Having bender spit stuff out to feed FC's reference methodology flow would be great but certainly limited in appeal at this point.

micprog:
Bender is generally designed to collect source files and provide these to the various tools. I don't know if it makes sense to provide the tools with physical related information, as there is currently no representation for this in bender. Adding this could be of interest, but I fear with varying technologies and parametrization of designs, even this physical information is more ideally suited to be handled outside of bender.

Nonetheless, we can add a custom script format for fusion-compiler. I do fear this can cause divergence; if the tools are similar and use the same commands for loading files, then improvements to one script format may get lost in the other. On the other hand, custom templates should now be an option, so this would be an alternative way to combine bender with additional outside information.

PR #164, opened 2024-04-13
Add error prop to generated vcs compile script

### LINKED ISSUE #163: VCS compile script doesn't propagate errors
Bender's script output in the vcs tera template emits a bunch separate parsing commands. Any individual command may fail but if the last command succeed the script is observed to succeed instead of stopping on an error.

uge:
Change the VCS compile script template to start with a 
```set -e```
to ensure error propagation to fix #163.

### DIFF
--- a/src/script_fmt/vcs_sh.tera
+++ b/src/script_fmt/vcs_sh.tera
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 # {{ HEADER_AUTOGEN }}
-
-# Set propagation of error to exit on first error
+{% if abort_on_error %}# Set propagation of error to exit on first error
 set -e
+{% endif %}
 ROOT="{{ root }}"
 {% if compilation_mode == 'separate' %}{% for group in srcs %}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/script_fmt/vcs_sh.tera [hunk 1]
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 # {{ HEADER_AUTOGEN }}
+
+# Set propagation of error to exit on first error
+set -e
```
micprog: I think masking this behind the existing `--no-abort-on-error` script option makes sense, for consistency.

```suggestion
{% if abort_on_error %}# Set propagation of error to exit on first error
set -e
{% endif %}
```

micprog (changes requested):

uge:
I wasn't familiar with that flag - thanks for the suggestion.

micprog (approved):
LGTM! :+1:

micprog (approved):
LGTM :+1:

PR #165, opened 2024-04-15
Fix assignment in genus template

cahz:
tcl assignment in genus does not allow for an equal sign

### DIFF
--- a/src/script_fmt/genus_tcl.tera
+++ b/src/script_fmt/genus_tcl.tera
@@ -2,7 +2,7 @@
 if [ info exists search_path ] {{ '{{' }}
   set search_path_initial $search_path
 {{ '}}' }}
-set ROOT = "{{ root }}"
+set ROOT "{{ root }}"
 {% if compilation_mode == 'separate' %}{% for group in srcs %}
 set search_path $search_path_initial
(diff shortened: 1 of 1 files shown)


src/script_fmt/genus_tcl.tera
```diff
@@ -1,8 +1,8 @@
 # {{ HEADER_AUTOGEN }}
-if [ info exists search_path ] {{ '{{' }}
+if [ info exists search_path ] {{ '{' }}
   set search_path_initial $search_path
-{{ '}}' }}
-set ROOT = "{{ root }}"
+{{ '}' }}
```
micprog: I am unsure if the change from two braces to one is needed, but I do not use genus, so I trust you tested this.

Another difference to the previous script prior to the use of templates, the `search_path_initial` variable used to be declared in an else statement as well, it might make sense to add this back: https://github.com/pulp-platform/bender/blob/25c518464b6a0c098e815c84f055e8fc2a4829f1/src/cmd/script.rs#L847-L852

cahz: The script fails with two braces, Genus then interprets it as a string.

The else branch for `search_path_initial` does not really matter. I have added it anyway.

micprog: Ok, thanks!

micprog (approved):
LGTM :+1:

PR #168, opened 2024-04-22
Put vcs and vsim defines in quotes

### LINKED ISSUE #162: Defines assumed to be string values, not numerical values
If a package declares a define:

```
  defines:
    CFG_W: 8
```
This results in a string being created for something like "bender script vcs":

`    +define+CFG_W="8" \
`
As a second concern, the +define is not generated with the "bender script flist" option (which I use as it provides the most flexibility for our needs), so there is no easy way to extract the defines that are specified.

SamuelRiedel:
Fixes #162


micprog (approved):
LGTM :+1:

Issue #172, opened 2024-05-10
Why does `bender` look upward for a `Bender.local` even if it is run in a Bender repo?


FrancescoConti:
In the following situation, `bender` works in a way I fail to rationalize. Assume you have a nest of IPs: `pulp` depends on `pulp_cluster` (and contains it), which depends on `cv32e40p` (and contains it).
Initally, you are working on `pulp_cluster` so you `bender clone` it and its dependency `cv32e40p`.
This results in two working dirs `pulp/working_dir/pulp_cluster` and `pulp/working_dir/cv32e40p`. The `Bender.local` in `pulp` gets updated.
So far so good.

When you finish working on `pulp_cluster` and `cv32e40p`, you push everything and update `pulp_cluster`'s `Bender.yml` - but you may want to push also `pulp_cluster`'s `Bender.lock`, e.g., for CI. That's when (in my view irrationally) a `bender update` will fail to work: it will actually point to the `Path` defined in the `pulp` `Bender.local`, even if the command is run inside `pulp_cluster`.

I understand why `bender` looks upwards in the file system hierarchy -- if you run it inside a nested folder in a repo, it makes sense to do that (like `git`). But it should stop at the first one it finds, not look further upwards -- what is the use case for that?

PR #173, opened 2024-05-10
Add warning about overrides used

micprog:
warns about #172


FrancescoConti (approved):
Makes sense to me. Although I personally prefer a git-like behavior (go upwards only until you encounter a folder containing a `Bender.yml` manifest, but no further).

PR #179, opened 2024-08-20
Fix mapping limitation in vendor

phsauter:
vendor did not handle it well when multiple files/directories were copied together into one (as is done in [croc](https://github.com/pulp-platform/croc/blob/ee968d10696f6081ef8feb82292cda7b2232cdbc/Bender.yml#L173)).

Fixing it involved prioritizing more specific mappings (eg file to file) over less specific ones (dir to dir) and excluding already diffed files when handling their parent directories.

Also fixes an annoying warning caused by a default configuration.

Tested `bender vendor init` and one random patch with `bender vendor patch` in opentitan_peripherals, register_interface, redundancy_cells and croc. The resulting directories were identical to upstream so it is unlikely to break compatibility with existing configurations.  
It should just make it possible to use more elaborate setups without issue.

### DIFF
--- a/src/cmd/vendor.rs
+++ b/src/cmd/vendor.rs
@@ -323,7 +323,7 @@ pub fn init(
 
     // Check if includes exist
     for path in vendor_package.include_from_upstream.clone() {
-        if !PathBuf::from(extend_paths(&[path.clone()], dep_path)?[0].clone()).exists() {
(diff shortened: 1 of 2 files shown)


niwis (approved):
Thanks a lot, that's a great improvement for mappings. LGTM

PR #183, opened 2024-09-06
Fix copying vendored files when they are symbolic links

alikates:
I'm working on a project where one vendored dependency uses symlinks for managing different SystemVerilog module versions. This was making bender's vendor commands to try to overwrite the symlink with the actual dereferenced file.

I assume this is not intended and the previous handling of symlinks was only for the special case where a vendored package has a symlink to another external directory. That's why I added another condition where the target of the symlink is an actual directory.

### DIFF
--- a/src/cmd/vendor.rs
+++ b/src/cmd/vendor.rs
@@ -757,6 +757,10 @@ pub fn copy_recursively(
         }
 
         let filetype = entry.file_type()?;
+        let canonical_path_filetype =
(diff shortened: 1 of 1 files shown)


alikates:
I just noticed I added an extra `canonicalize` for the `orig` path, which should not be needed. I will fix that together with the formatting.

micprog (approved):
LGTM :+1:

PR #188, opened 2024-11-18
adding flake.nix and .lock for building with nix flakes

haflex:
Enable building of bender on nix flake enabled systems

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,77 @@
+{
+  "nodes": {
+    "crane": {
+      "locked": {
+        "lastModified": 1731098351,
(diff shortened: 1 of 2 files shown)


micprog (approved):
I can add these changes but will not maintain them, as I am not familiar with and do not use nix flakes.

haflex:
As long as nothing drastically changes in the repo there shouldn't be any need to do anything. And it won't hurt anyone, right?
I don't know what your policies are, if you need a maintainer? If so, I can do that. At least when a new version comes out I can check that it still bulids with nix.

PR #192, opened 2025-03-17
git: handle the case where `git tag` opens an editor

midnightveil:
For me, the `git tag bender-tmp-<commit> <commit> --force` opens a nano process, which then causes the `bender checkout` to just hang wait for git to complete.

Add a message just to make it go on its merry way.

(An alternate solution would probably be to add `--no-sign`, as this is the cause in my specific case).

cc @niwis


micprog (changes requested):
As these are temporary tags, I'd prefer to go the `--no-sign` route, as a message will imply an annotated tag (see https://git-scm.com/docs/git-tag).

micprog (approved):
LGTM, thanks!

PR #193, opened 2025-03-20
Support for Cadence Xcelium simulator

uge:
Adds basic support for Cadence Xcelium simulator.

Only verilog (not vhdl) tested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,6 +290,7 @@ Individual commands may also set tool-specific targets:
 
 - `vsim`
 - `vcs`
+- `xm`
 - `verilator`
 - `synopsys`
 - `riviera`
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -290,6 +290,7 @@ Individual commands may also set tool-specific targets:
 
 - `vsim`
 - `vcs`
+- `xm`
```
micprog, line 293: Can we change this name to `xcelium`? This applies to all uses below (command name, docs, ...)

src/script_fmt/xm_sh.tera
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+# {{ HEADER_AUTOGEN }}
+{% if abort_on_error %}# Set propagation of error to exit on first error
+set -e
+{% endif %}
+ROOT="{{ root }}"
+{% for group in srcs %}
+{% if group.file_type == 'verilog' %}xmvlog  -sv \
```
micprog: I see you have support for verilog files, which is a great start. As bender is generally built to support vhdl files as well, and this simulator supports vhdl, it would be great to ensure compatibility here as well.

README.md
uge: Done

src/script_fmt/xm_sh.tera
uge: We don't have any VHDL code internally to test with. If there is a public Bender repo with VHDL code and a simulation target, please refer me to it an I'll try to include the template.

yvantor: Hi @uge, we have a [can_bus](https://github.com/AlSaqr-platform/can_bus) entirely written in VHDL. You can also find it integrated in [Cheshire](https://github.com/pulp-platform/cheshire/tree/yt/can) if you need an example with both Verilog and VHDL IPs.

PR #194, opened 2025-03-20
Add .e as a variant of verilog for encrypted verilog

uge:
Some IP vendors distribute encrypted verilog IP as .e extension


micprog:
I wanted to do some restructuring within Bender and the manifest to allow users to specify the type of each individual file from within the Bender.yml (even allowing to override the default recognized .v, .vhd, ...). I don't think .e will be uniquely identifiable as encrypted verilog.

PR #197, opened 2025-04-01
Refactor update command and add new features

mrogenmoserTT:
### Fixed
- Fix `update` failing pull of inexistent commit when multiple repositories are indicated.
- Fix checkout for commit not yet fetched.

### Added
- `update`: Update repositories in checkout dir if conditions match:
  - Folder is a git repository
  - Git state is clean
  - `--ignore-checkout-dir` flag to update IPs ignoring the state inside the checkout directory.
- `update`: Tell user the lockfile version when solving a conflict.
- Print dependency updates executed.
- Enable updating of individual dependencies (and recursive dependencies if desired).

### Changed
- `update`: Clean up alignment of manual resolution output.
- `checkout`: When using `checkout_dir`, overwrite existing dependencies if not changed, warning if not checked out, flag to force checkout.
- `update`: Update `checkout_dir` if no internal changes.
- Execute checkout instead of clone to checkout correct dependency versions when reasonable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Fixed
+- Fix `update` failing pull of inexistent commit when multiple repositories are indicated.
 
-# 0.28.2 - 2025-03-31
+### Added
+- `update`: Update repositories in checkout dir if conditions match:
+  - Folder is a git repository
+  - Git state is clean
+  - Commit matches with lockfile
+  - `--ignore-checkout-dir` flag to update IPs ignoring the state inside the checkout directory.
+- `update`: Tell user the lockfile version when solving a conflict.
+
+## 0.28.2 - 2025-03-31
 ### Fixed
 - Put `vcs`, `vsim`, and `riviera` defines in quotes.
 - Fix `genus` script initialization.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
   - `--ignore-checkout-dir` flag to update IPs ignoring the state inside the checkout directory.
 - `update`: Tell user the lockfile version when solving a conflict.
 
+### Changed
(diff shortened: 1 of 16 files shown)


mrogenmoserTT:
Hi @fischeti,
Thanks for the feedback!
> - If I want to update a single dependency e.g. `bender update common_cells`, it will update the `Bender.lock` file but not `Bender.yml`. So potentially, the `Bender.yml` gets out of date. But I don't have a better solution, because I guess modifying the `Bender.yml` automatically is also not a good option.

Correct, the `Bender.yml` does not get updated. However, the `Bender.yml` indicates requirements that are respected by the `update` command, which will pick the newest compatible version available. So, the version requirement set in the `Bender.yml` is still respected, meaning that while it may point to an older version, the requirement is not necessarily out of date.

> - The --ignore-checkout-dir name is a bit confusing, since it implies that the checkout_dir setting is ignored, but actually it overwrites the checkout dir.

This is an update to the behavior when specifying a `checkout_dir` in the `Bender.yml`. Previously, bender enforced the use of this directory, no longer updating any dependency checked out within. Now, it will continue updating the directory unless it is dirty or no longer a git folder, which can be overridden with the flag. This does not affect the behavior when the `checkout_dir` is not set (i.e., checkouts in the `.bender/git/checkouts/` directory).

fischeti:
> Hi @fischeti, Thanks for the feedback!
> 
> > * If I want to update a single dependency e.g. `bender update common_cells`, it will update the `Bender.lock` file but not `Bender.yml`. So potentially, the `Bender.yml` gets out of date. But I don't have a better solution, because I guess modifying the `Bender.yml` automatically is also not a good option.
> 
> Correct, the `Bender.yml` does not get updated. However, the `Bender.yml` indicates requirements that are respected by the `update` command, which will pick the newest compatible version available. So, the version requirement set in the `Bender.yml` is still respected, meaning that while it may point to an older version, the requirement is not necessarily out of date.
> 

Ah I see, that makes more sense 👍 

> > * The --ignore-checkout-dir name is a bit confusing, since it implies that the checkout_dir setting is ignored, but actually it overwrites the checkout dir.
> 
> This is an update to the behavior when specifying a `checkout_dir` in the `Bender.yml`. Previously, bender enforced the use of this directory, no longer updating any dependency checked out within. Now, it will continue updating the directory unless it is dirty or no longer a git folder, which can be overridden with the flag. This does not affect the behavior when the `checkout_dir` is not set (i.e., checkouts in the `.bender/git/checkouts/` directory).

Okay, I think it's fine. This feature is anway not really used a lot.

fischeti (approved):

PR #203, opened 2025-05-01
Add support for glob in manifest source files.

mrogenmoserTT:


fischeti (approved):
LGTM! Maybe adding an example in the README would make sense

PR #206, opened 2025-05-01
Improve script command output

mrogenmoserTT:
- Fix incdirs in flist-plus script format
- Allow non-caps defines


fischeti (approved):
LGTM!

PR #207, opened 2025-05-01
Add support for verilog/vhdl files with custom file endings

micprog:
This PR adds support for custom file endings of vhdl, verilog, or systemverilog files. This is relevant for scripts generated directly for custom tools, e.g., to import encrypted files that have endings such as `.svp`, `.vp`, or `.e` (or anything else you could think of). Custom file endings can now be allocated a file type as follows:

```yaml
  # Source files can have custom fileendings
  - sv: vendor/encrypted_sv_src.svp
  - v: vendor/encrypted_v_src.vp
  - vhd: vendor/encrypted_vhd_src.e
```

### DIFF
--- a/src/src.rs
+++ b/src/src.rs
@@ -295,7 +295,14 @@ pub enum SourceFile<'ctx> {
 impl<'ctx> fmt::Debug for SourceFile<'ctx> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
-            SourceFile::File(path, _) => fmt::Debug::fmt(path, f),
+            SourceFile::File(path, ty) => {
+                fmt::Debug::fmt(path, f)?;
+                if let Some(t) = ty {
+                    write!(f, " as {:?}", t)
+                } else {
+                    write!(f, " as <unknown>")
+                }
+            }
             SourceFile::Group(ref srcs) => fmt::Debug::fmt(srcs, f),
         }
     }
--- a/src/src.rs
+++ b/src/src.rs
@@ -232,7 +232,7 @@ impl<'ctx> SourceGroup<'ctx> {
         };
         for file in subfiles {
             match file {
-                SourceFile::File(_) => {
+                SourceFile::File(_, _) => {
                     files.push(file);
                 }
                 SourceFile::Group(grp) => {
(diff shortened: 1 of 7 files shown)


micprog:
Replaces assumptions proposed in #194.

src/src.rs
```diff
@@ -270,21 +270,39 @@ impl<'ctx> SourceGroup<'ctx> {
     }
 }
 
+/// File types for a source file.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum SourceType {
+    /// A Verilog file.
+    Verilog,
+    // /// A SystemVerilog file.
+    // SystemVerilog,
```
fischeti, line 354: Can be removed?

fischeti (commented):
LGTM! The only suggestion I have is that the new keys `sv`, `v` and `vhd` could be nested instead of being equal to `files`. For instance, something like this would make a bit more sense to me:

```yml
sources:
  # Individual source files are simple string entries:
  - src/package.sv
  - src/file1.vhd
  - src/file2.vhd

  # Source files can be grouped:
  - files:
      - src/stuff/pkg.sv
      - src/stuff/top.sv
      - sv: vendor/encrypted_sv_src.svp
      - v: vendor/encrypted_v_src.vp
      - vhd: vendor/encrypted_vhd_src.e
```

micprog:
> LGTM! The only suggestion I have is that the new keys `sv`, `v` and `vhd` could be nested instead of being equal to `files`. For instance, something like this would make a bit more sense to me:
> 
> ```yaml
> sources:
>   # Individual source files are simple string entries:
>   - src/package.sv
>   - src/file1.vhd
>   - src/file2.vhd
> 
>   # Source files can be grouped:
>   - files:
>       - src/stuff/pkg.sv
>       - src/stuff/top.sv
>       - sv: vendor/encrypted_sv_src.svp
>       - v: vendor/encrypted_v_src.vp
>       - vhd: vendor/encrypted_vhd_src.e
> ```

This should already be supported. The sv/v/vhd keys can be used in place of a sourcefile (or a sourcegroup), so even within sourcegroups we can use these keys. These sourcegroups and sourcefiles can be nested hierarchically as desired. Let me know if this doesn't work for you, or do you want to exclude the option of providing a format-specific file at the top hierarchy?

fischeti (approved):

PR #210, opened 2025-05-20
Multi-Arch Releases

Xeratec:
This PR adds support for building multi-arch releases. See https://github.com/Xeratec/bender/releases/tag/v0.28.3-rc1 for an example.

## Added
- New `release_arm64` build stage

## Changes
- Renamed `release` stage to `release_amd64`

## Fixed
- Fixed `ENV` warning in Dockerfile

## Note
- During the build, the release files are placed in `target/<architecture>/<platform>` (e.g. `target/arm64/ubuntu22.04`)
- To keep the release names compatible with earlier releases, `amd64` is replaced with `x86_64`

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -75,7 +75,7 @@ jobs:
         run: |
           export tgtname=$(echo ${{ matrix.os }} | tr -d ':')
           export platform=$(echo ${{ matrix.platform }} | awk -F'/' '{print $NF}')
-          .github/scripts/package.sh $tgtname $platform;
+          .github/scripts/package.sh $platform $tgtname;
         shell: bash
       - name: Upload Release Asset
         uses: softprops/action-gh-release@v1
@@ -120,7 +120,7 @@ jobs:
         run: |
           export tgtname=$(echo ${{ matrix.os }} | tr -d ':')
           export platform=$(echo ${{ matrix.platform }} | awk -F'/' '{print $NF}')
-          .github/scripts/package.sh $tgtname $platform;
+          .github/scripts/package.sh $platform $tgtname;
         shell: bash
       - name: Upload Release Asset
         uses: softprops/action-gh-release@v1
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yaml
```diff
@@ -80,8 +131,6 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   release-gnu:
     runs-on: ubuntu-latest
-    strategy: 
-      fail-fast: false
```
micprog: Any reason you're removing this?

.github/workflows/release.yaml
```diff
@@ -126,8 +175,6 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   release-macos:
     runs-on: macos-latest
-    strategy:
-      fail-fast: true
```
micprog, line 130: Any reason you're removing this?

.github/workflows/release.yaml
```diff
@@ -80,8 +131,6 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   release-gnu:
```
micprog: If we support arm64, can we add a non-os-specific linux release for this as well? This container should make it possible: https://quay.io/repository/pypa/manylinux2014_aarch64

micprog (commented):
Thanks for this extension! I have a few comments in the review. Also, please check if the names chosen work with the download script here: 
```
curl --proto '=https' --tlsv1.2 https://pulp-platform.github.io/bender/init -sSf | sh
```
source: https://github.com/pulp-platform/bender/blob/website/init
If the script doesn't support arm64/aarch64 yet, extending it would be great!

.github/workflows/release.yaml
Xeratec: Yes, because it is invalid syntax without the matrix keyword.

.github/workflows/release.yaml
Xeratec: Yes, because it is invalid syntax without the matrix keyword.

.github/workflows/release.yaml
Xeratec: Sure, I can try it out.

Xeratec: Done, let me know if you want other changes. See https://github.com/Xeratec/bender/releases/tag/v0.28.3-rc2

micprog (approved):
LGTM, thanks! :+1:

Issue #211, opened 2025-05-20
ARM64 Download Script Support


Xeratec:
Thanks for this extension! I have a few comments in the review. Also, please check if the names chosen work with the download script here: 
```
curl --proto '=https' --tlsv1.2 https://pulp-platform.github.io/bender/init -sSf | sh
```
source: https://github.com/pulp-platform/bender/blob/website/init
If the script doesn't support arm64/aarch64 yet, extending it would be great!

_Originally posted by @micprog in https://github.com/pulp-platform/bender/pull/210#pullrequestreview-2854217824_

PR #212, opened 2025-05-21
ARM64 Download Script Support

Xeratec:
This PR adapts the download script to support ARM64 builds, as described in #211.

Note: I did not test it on Windows, but as this part of the code is untouched, I argue this is unnecessary.

## Added
- Added logic to handle `arm64-linux-gnu` platforms

## Fixed
- Fixed some syntax to work with `sh` (previously only worked bith `bash`)

## ToDo
- [x] Test script for new `arm64-linux-gnu-ubuntu22.04` platform
- [x] Test script for new `arm64-linux-gnu-ubuntu16.04` platform
- [x] Test script for new `arm64-linux-gnu-debian12` platform
- [x] Test script for new `amd64-linux-gnu-ubuntu22.04` platform
- [x] Test script for new `amd64-linux-gnu-ubuntu16.04` platform
- [x] Test script for new `amd64-linux-gnu-debian12` platform
- [x] Test script for new `macos-latest` platform

### DIFF
--- a/init
+++ b/init
@@ -49,12 +49,16 @@ main() {
         local _arch="universal-apple-darwin"
     fi
 
-    if [ "$_arch" != "x86_64-linux-gnu" ] && [ "$_arch" != "universal-apple-darwin" ]; then
+    if [ "$_arch" = "aarch64-linux-gnu" ]; then
+        local _arch="arm64-linux-gnu"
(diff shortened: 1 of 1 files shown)


micprog (approved):
LGTM :+1:

PR #219, opened 2025-08-28
Add Xcelium script format and --incdir flag

DanielKellerM:
This PR adds support for the Cadence Xcelium simulator to Bender.

### Changes

**New command: `bender script xcelium`**
- Generates a flat `.f` file with per-group incdirs, defines, and source files
- Consistent with how `vsim` and `vcs` templates work (no `-makelib`/`-endlib` library scoping)
- Supports `--relative-path` for `$ROOT`-relative output
- Includes `source_annotations` metadata comments (opt-out with `--no-source-annotations`)

**New global flag: `--incdir` / `-I`**
- Injects user-provided include directories into both the global incdir list and each per-source-group's incdirs
- Applies to all script formats

**Struct changes**
- Added `name: Option<String>` to `TplSrcStruct` (populated from `src.package`)
- Added `root_package` to template context

### Testing

Verified end-to-end with a 5,107-module Snitch Cluster design:
- Flat single-pass compile+elaborate: **14.7s wall, 1 xrun invocation**
- Both flat and library-scoped approaches produce identical elaboration results

### DIFF
--- a/src/cmd/script.rs
+++ b/src/cmd/script.rs
@@ -454,7 +454,6 @@ fn emit_template(
 ) -> Result<()> {
     tera_context.insert("HEADER_AUTOGEN", HEADER_AUTOGEN);
     tera_context.insert("root", sess.root);
-    tera_context.insert("root_package", &sess.manifest.package.name);
     // tera_context.insert("srcs", &srcs);
     tera_context.insert("abort_on_error", &!args.no_abort_on_error);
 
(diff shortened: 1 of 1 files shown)


yvantor:
@micprog I am jumping in just to mention that I worked as a beta tester for this and seems to be working pretty well. I am currently testing it also in Cheshire.

yvantor:
@DanielKellerM do you have bandwidth to look into this?

PR #220, opened 2025-09-11
Add support to parse a filelist specified within a Bender.yml

mrogenmoserTT:


mrogenmoserTT:
The most recent push should address 1,2,4, and 6. For the paths, right now this only supports files relative to the file list file path, as making these file lists relative to yet another directory would require more information to be passed. Properly relativizing the path still needs to be addressed.

PR #221, opened 2025-09-11
Number warnings (and some errors) and enable suppressing of the warnings

mrogenmoserTT:


fischeti (approved):
LGTM, I tried it out and the functionality is as expected.

One optional improvement would be to somehow enumerate the warnings automatically or specify the list of possible warnings somewhere. Otherwise, if future warnings are added, one needs to search for the currently highest warning number.

src/cli.rs
```diff
@@ -95,6 +104,19 @@ pub fn main() -> Result<()> {
...
 
+    let suppressed_warnings: IndexSet<String> = matches
+        .get_many::<String>("suppress")
+        .unwrap_or_default()
+        .map(|s| s.to_owned())
+        .collect();
+
+    let suppressed_warnings: IndexSet<String> =
+        if suppressed_warnings.contains("all") || suppressed_warnings.contains("Wall") {
+            (1..23).map(|i| format!("W{:02}", i)).collect()
```
micprog, line 115: @fischeti The possible warning IDs are indicated here and will need to be incremented when adding a warning.

PR #225, opened 2025-10-06
Update clone command, fail cloning of path deps

### LINKED ISSUE #208: `bender clone` on Path Subdependencies Fails
If a Bender repo `A` contains a subdependency `B`, specified in its `Bender.yml` as a `Path` dependency that is located in a subdirectory of the repo with its own `Bender.yml`, `bender clone B -p my_local_stuff` fails with the following error message:
```
error: remote source already exists.
error: git renaming remote origin failed
```
I assume this is because Bender assumes that `B` is also a git repo, but it isn't.
My proposed solution would be to treat `Path` subdependencies slightly differently:
- `bender clone A` could update `Bender.lock` (and `Bender.local`?) not just for the `A` dependency but also for all `Path` subdependencies of `A`, pointing them to the `clone`'d local version contained in the `clone` of A.
- `bender clone B` could fail and tell the user that `B` is a `Path` subdependency in  `A`, and they should `bender clone A` instead.

micprog:
Fixes #208


src/cmd/clone.rs
```diff
@@ -57,6 +56,17 @@ pub fn run(sess: &Session, path: &Path, matches: &ArgMatches) -> Result<()> {
         }
     }
 
+    // Check if dependency is a git dependency
+    match sess.dependency_source(depref) {
+        DependencySource::Git { .. } | DependencySource::Registry => {}
+        DependencySource::Path { .. } => {
+            Err(Error::new(format!(
+                "Dependency `{}` is not a git dependency, cannot clone.",
+                dep
```
phsauter, line 67: I would consider a more instructive error message, I can see people getting very frustrated at this especially if they are new to Bender.

src/cmd/clone.rs
```diff
@@ -179,7 +176,7 @@ pub fn run(sess: &Session, path: &Path, matches: &ArgMatches) -> Result<()> {
         let mut new_str = String::new();
         if local_file_str.contains("overrides:") {
             let split = local_file_str.split('\n');
-            let test = split.clone().last().unwrap().is_empty();
+            let test = split.clone().next_back().unwrap().is_empty();
```
phsauter: I think you can use this:
https://doc.rust-lang.org/std/string/struct.String.html#method.ends_with

fischeti (commented):
I tried it out on an example that is similar to what is described in https://github.com/pulp-platform/bender/issues/208. However, I get the following behaviour:

1. `bender clone A` still works, and it does not update the path of `B` in Bender.lock i.e. it still points to the one in `.bender`. Which I think is fine.
1. `bender clone B` throws an error, but not the new one I would expect i.e. I get a `fatal: '".bender/path/to/path_dep"' does not appear to be a git repository`

Shouldn't the second case throw the new error`Dependency is not a git dependency, cannot clone.`?

src/cmd/clone.rs
micprog: Is 
```suggestion
            Err(Error::new(format!(
                "Dependency `{}` is a path dependency. `clone` is only implemented for git dependencies.",
                dep
```
more instructive?

phsauter: Yes, this should give people enough info to figure out their problem.

PR #228, opened 2025-11-06
Add target defines to sources output

micprog:


Lore0599 (approved):
Thanks for the added feature, I think this could be very helpful in the future :)

PR #229, opened 2025-11-07
Add warnings for nonexistend include directory and errors for missing path dependency

mrogenmoserTT:


mrogenmoserTT:
I adjusted the internals to only error if the missing include_directory is actually required :+1:

micprog:
As far as I understand, there is no inherent harm in including a non-existent directory for any downstream tool, although it is bad practice, unlike if a file doesn't exist. Therefore, I think it makes the most sense to convert this error into a warning, not an error.

PR #230, opened 2025-11-07
Add snapshot command

mrogenmoserTT:


src/cmd/snapshot.rs
```diff
@@ -0,0 +1,195 @@
...
+    if local_path.exists() && !snapshot_list.is_empty() {
+        let local_file_str = match std::fs::read_to_string(&local_path) {
+            Err(why) => Err(Error::new(format!(
+                "Reading Bender.local failed with msg:\n\t{}",
+                why
+            )))?,
+            Ok(local_file_str) => local_file_str,
+        };
+        let mut new_str = String::new();
+        if local_file_str.contains("overrides:") {
```
fischeti, line 128: Not sure but maybe it would be more straightforward to deserialize and serialize the YAML file again instead of parsing it.

src/cmd/snapshot.rs
```diff
@@ -0,0 +1,195 @@
...
+            let split = local_file_str.split('\n');
+            let test = split.clone().next_back().unwrap().is_empty();
+            for i in split {
+                for (name, _, _) in &snapshot_list {
+                    if i.contains(name) {
+                        new_str.push('#');
+                    }
+                }
```
fischeti, line 136: What's the idea behind commenting out the old version?

fischeti (commented):
I really like this feature! thanks for putting in the effort. I have some considerations below:

First, the dependencies to snapshot are currently based on the `working_dir`, but wouldn't it make more sense to specify the actual dependency to snapshot instead of the entire working directory?

I was also wondering what the purpose is of modifying the `Bender.local` file with the same commit hash that is now part of the `Bender.lock`. I think it could make more sense to keep the path dependency as is (in case you would want to continue working on it locally) or completely remove it from the `Bender.local` to let the dependency be handled like a normal one. This could also be handled by an additional flag. Maybe I am also missing something with the already existing flags.

I tried out the feature myself and it works on my machine😉 Maybe a bit of documentation in the README would help the adoption of this nice feature.

src/cmd/snapshot.rs
mrogenmoserTT: This is not done to preserve comments in the Bender.yml file, however properly handling this will likely prevent errors.

src/cmd/snapshot.rs
mrogenmoserTT: Only a single version should be specified in the Bender.local. The previous version is commented to keep a record. This is similar behavior to the one implemented for the `clone` command.

fischeti:
> The goal of this command was to enable sharing of a reproducible setup without sharing the local clones of dependencies that were being worked on. To enable this, all dependencies should be re-linked to the corresponding git projects they originate from. While it is possible to specify which projects you would like to link back, I think that way things can get lost more easily that way as you may forget additional dependencies you were working on. Of course, we can look at optionally adding this in the future to limit the attacked scope.

I see makes sense. Specifying a single dependency would be a nice to have, but we can also leave it for the future.

PR #231, opened 2025-11-20
Fix print behavior

mrogenmoserTT:
- Ensure informative print is not in stdout (e.g. fix for scripts)
- Use write to stdout instead of print to enable piping bender output into another command
- Remove compile warnings for release


phsauter (approved):
Looks good to me

PR #235, opened 2025-12-04
Add functionality to pass targets to a dependency

mrogenmoserTT:
Add `pass_targets` to dependencies to allow passing targets for hierarchical file filtering (not only global).

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ dependencies:
   common_verification: { git: "git@github.com:pulp-platform/common_verification.git", version: "0.1" }
 
   # Git revision dependency.
-  common_cells: { git: "git@github.com:pulp-platform/common_cells.git", rev: master }
+  common_cells: { git: "git@github.com:pulp-platform/common_cells.git", rev: master, pass_targets: ["CC_CUSTOM_TARGET"] }
 
 # Freeze any dependency updates. Optional. False if omitted.
 # Useful for chip packages. Once the chip is in final tapeout mode, and
--- a/src/cmd/parents.rs
+++ b/src/cmd/parents.rs
@@ -10,6 +10,7 @@ use indexmap::IndexMap;
 use tabwriter::TabWriter;
 use tokio::runtime::Runtime;
 
+use crate::config::Dependency;
 use crate::error::*;
 use crate::sess::{DependencyConstraint, DependencySource};
 use crate::sess::{Session, SessionIo};
@@ -18,11 +19,18 @@ use crate::sess::{Session, SessionIo};
 pub fn new() -> Command {
     Command::new("parents")
         .about("List packages calling this dependency")
+        .alias("parent")
         .arg(
             Arg::new("name")
                 .required(true)
                 .num_args(1)
                 .help("Package names to get the parents for"),
+        )
+        .arg(
+            Arg::new("targets")
+                .long("targets")
+                .num_args(0)
+                .help("Print the passed targets to the dependency"),
         )
 }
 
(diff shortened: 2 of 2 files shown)


src/config.rs
```diff
@@ -616,6 +634,12 @@ impl Validate for PartialDependency {
         pre_output: bool,
         suppress_warnings: &IndexSet<String>,
     ) -> Result<Dependency> {
+        let pass_targets = self
+            .pass_targets
+            .unwrap_or_default()
+            .into_iter()
+            .map(|s| s.to_lowercase())
```
fischeti, line 641: Are all passed targets internally converted to lowercase?

src/sess.rs
```diff
@@ -1439,6 +1449,78 @@ impl<'io, 'sess: 'io, 'ctx: 'sess> SessionIo<'sess, 'ctx> {
...
+                    ),
+                );
+            }
+        }
+
+        let final_indexmap: IndexMap<String, TargetSet> = final_indexmap
+            .into_iter()
+            .map(|(k, v)| (k, TargetSet::new(v.into_iter())))
+            .collect();
+
```
fischeti, line 1523: I had to study this for a while until I understood the concepts of ranks in dependency. I think the implementation is correct. I was wondering though whether the targets need to be reduced hierarchically by rank or whether it is also possible to just flatten all the manifests and then reduce/collect the targets. This would be much easier to implement since it just a nested for loop over manifests + dependency of a manifest. Or do you see an issue with that? I am not sure whether that might cause issues with ordering etc.,

src/resolver.rs
```diff
@@ -354,7 +358,7 @@ impl<'ctx> DependencyResolver<'ctx> {
                 let name = name.as_str();
                 debugln!("resolve: registering {} from lockfile", &name);
                 let dep = match &locked_package.source {
-                    LockedSource::Path(p) => config::Dependency::Path(p.clone()),
+                    LockedSource::Path(p) => config::Dependency::Path(p.clone(), Vec::new()),
```
fischeti, line 362: I am not sure, but would it make sense to record the targets in the lockfile?

src/cmd/script.rs
```diff
@@ -271,6 +278,7 @@ pub fn run(sess: &Session, matches: &ArgMatches) -> Result<()> {
             files: Default::default(),
             dependencies: Default::default(),
             version: None,
+            passed_targets: TargetSet::empty(),
         });
```
fischeti, line 282: Unrelated to your changes, but we could derive/implement the `Default` trait for `SourceGroup`🤓 Because there are a couple of locations where an empty source group is created.

fischeti (approved):
LGTM!

src/config.rs
mrogenmoserTT: Targets are handled as case-insensitive in bender (especially as the resulting defines are capitalized).

src/sess.rs
mrogenmoserTT: We want to ensure the output for the scripts is ordered by ranks to have dependencies compiled before the top by any tool that uses it. I'm not sure of the effects in this specific location, but keeping this order can't hurt. There are likely quite a few optimizations that could be done around bender for this.

src/resolver.rs
mrogenmoserTT: The lockfile is only changed when a `bender update` is run, but realistically, the targets that are passed to dependencies could be adjusted irrespective of full-blown dependency updates. Thus, I think that is an unhelpful restriction to keep this information in the lockfile.

src/cmd/script.rs
mrogenmoserTT: That sounds like a great idea! Separate PR?

micprog: #248

PR #237, opened 2025-12-11
Simplify choices in update conflict resolution

micprog:


src/util.rs
```diff
@@ -214,3 +217,290 @@ pub fn try_modification_time<P: AsRef<Path>>(path: P) -> Option<SystemTime> {
...
+                    } else {
+                        0
+                    },
+                    pre: semver::Prerelease::EMPTY,
+                    build: semver::BuildMetadata::EMPTY,
+                };
+                if top_bound > max_exact {
+                    found = true;
+                    top_bound = max_exact;
+                }
```
fischeti, line 254: I think this can be moved out the match statement, since it is the same for every branch.

src/util.rs
```diff
@@ -214,3 +217,290 @@ pub fn try_modification_time<P: AsRef<Path>>(path: P) -> Option<SystemTime> {
...
+                    patch: comp.patch.unwrap_or(0),
+                    pre: semver::Prerelease::EMPTY,
+                    build: semver::BuildMetadata::EMPTY,
+                };
+                if top_bound > max_less {
+                    found = true;
+                    top_bound = max_less;
+                }
+            }
+            semver::Op::LessEq => {
```
fischeti: Isn't `LessEq` the same as `Equal`?

src/util.rs
```diff
@@ -214,3 +217,290 @@ pub fn try_modification_time<P: AsRef<Path>>(path: P) -> Option<SystemTime> {
...
+                    top_bound = max_less_eq;
+                }
+            }
+            semver::Op::Tilde => {
+                let max_tilde = Version {
+                    major: if comp.minor.is_some() {
+                        comp.major
+                    } else {
+                        comp.major + 1
+                    },
```
fischeti: I believe `major` should stay the same:

from docs:

```
 ~I.J.K — equivalent to >=I.J.K, <I.(J+1).0
```

src/resolver.rs
```diff
@@ -718,14 +728,64 @@ impl<'ctx> DependencyResolver<'ctx> {
...
+                            version_req_top_bound(vb)
+                                .unwrap_or(Some(semver::Version::new(u64::MAX, u64::MAX, u64::MAX)))
+                                .cmp(&version_req_top_bound(va).unwrap_or(Some(
+                                    semver::Version::new(u64::MAX, u64::MAX, u64::MAX),
+                                )))
+                        }
+                        (
+                            DependencyConstraint::Revision(ra),
+                            DependencyConstraint::Revision(rb),
+                        ) => ra.cmp(rb),
```
fischeti, line 771: revisions are alphabetically ordered?

src/util.rs
micprog: The `max_exact` only lives in the match cases.

src/resolver.rs
micprog: Yes. This ensures consistency in the printed list.

src/util.rs
micprog: Yes, the bottom bound can also be simplified a lot. I pushed an update.

PR #239, opened 2026-01-01
Add CLI progress bars

fischeti:
# CLI Progress bars

Adds progress bars to bender actions that require longer network access. A taste is shown below*

bender

*some style changes have been done since this recording.

## Implementation

The progress bars are rendered with the crate [`indicatif`](https://github.com/console-rs/indicatif), that allows rendering of multiple progress bars at once. The state of the progress bars is extracted from the stderr output of the spawned git commands i.e. a separate task is spawned alongside git tasks to parse git output lines and update the progress bar state. I initially thought about adopting the [`git2`](https://github.com/rust-lang/git2-rs) crate, which would have built in progress callbacks. However, this would have been a larger change and also poses some challenges since those git tasks are not asynchronous and also credential handling would be more complicated, which is why I opted for stderr parsing in the end.

## Related Changes

Apart from the implementation of the progress bars, some additional changes were necessary:

### `error` macros

Previously, all messages were simply printed to stdout/stderr with `println!/eprintln!` without any specific order. Rendering the progress bars together with warning/error messaged caused unwanted artifacts. The reason is that the progress bar requires exclusive access over stdout/stderr, since it needs to clear lines again. If warning/error messages are sent in between, it might clear the warning/error message instead of the actual progress bar which is left behind as an artifact. For that matter, the `MultiProgress` instance can be suspended, which essentially means clearing all progress bars, writing the desired custom message to stdout/stderr and then redraw all the progress bars. This means the previous `warnln!`, `errorln!` etc., macros are now passed through the `MultiProgress.suspend()` function.

### `git` spawns

* Where appropriate the git `spawn` functions now accept a `Option<ProgressHandler>` argument. If `None` is passed, the behaviour is as before, whereas `Some(pb)` will add a new progress bar.
* The `git submodule update` commands are only performed if a dependency contains a `.gitmodules`. Otherwise, too many no-op progress messages are spawned.

## TODOs

- [x] Check more bender commands. I focused mostly on `checkout`, but others use the same underlying functions so there should not be an issue in theory.
- [x] Check output in log files (e.g. CI). In theory, this should work outside of the box.
- [x] Revert git throttling changes

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -232,19 +232,17 @@ impl<'ctx> Git<'ctx> {
 
     /// Fetch the tags and refs of a remote.
     pub async fn fetch(self, remote: &str, pb: Option<ProgressHandler>) -> Result<()> {
-        let r1 = String::from(remote);
-        let r2 = String::from(remote);
         self.clone()
             .spawn_with(
-                |c| c.arg("fetch").arg("--prune").arg(r1).arg("--progress"),
+                |c| {
+                    c.arg("fetch")
+                        .arg("--tags")
+                        .arg("--prune")
+                        .arg(remote)
+                        .arg("--progress")
+                },
                 pb,
             )
-            .and_then(|_| {
-                self.spawn_with(
-                    |c| c.arg("fetch").arg("--tags").arg("--prune").arg(r2),
-                    None,
-                )
-            })
             .await
             .map(|_| ())
     }


src/error.rs
```diff
@@ -53,20 +61,20 @@ macro_rules! diagnostic {
 #[derive(PartialEq, Eq)]
 pub enum Severity {
     Debug,
-    Note,
-    Warning,
```
micprog, line 57: Just so I properly understand - is the warning severity now removed, or where is this handled?

log.txt
```diff
@@ -0,0 +1,10 @@
+    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.25s
```
micprog: Is this file needed?

src/cmd/vendor.rs
```diff
@@ -416,50 +434,52 @@ pub fn apply_patches(
...
-
-                        // limit to specific file for file links
-                        if is_file {
-                            let file_path = patch_link.from_prefix.to_str().unwrap();
-                            c.arg("--include").arg(file_path);
-                        }
-
-                        c
+                future::lazy(|_| Ok(()))
+                    .and_then(|_| {
```
micprog: I believe this part can be removed

src/error.rs
```diff
@@ -138,16 +146,3 @@ impl From<std::io::Error> for Error {
...
-macro_rules! stageln {
-    ($stage:expr, $($arg:tt)*) => {
-        $crate::error::println_stage($stage, &format!($($arg)*))
-    }
-}
-
-/// Print stage progress.
-pub fn println_stage(stage: &str, message: &str) {
-    eprintln!("\x1B[32;1m{:>12}\x1B[0m {}", stage, message);
-}
```
micprog, line 153: Removing the `{:>12}` from the `stageln!` output removes the existing right alignment of the stage part of the output, leading to a less clean console. Can we investigate adding this back?

src/git.rs
```diff
@@ -158,26 +200,42 @@ impl<'ctx> Git<'ctx> {
...
-            .and_then(|_| self.spawn_with(|c| c.arg("fetch").arg("--tags").arg("--prune").arg(r2)))
+            .spawn_with(
+                |c| c.arg("fetch").arg("--prune").arg(r1).arg("--progress"),
+                pb,
+            )
+            .and_then(|_| {
+                self.spawn_with(
+                    |c| c.arg("fetch").arg("--tags").arg("--prune").arg(r2),
+                    None,
+                )
```
micprog: Am I seeing correctly that we only mark the first part of fetching with a progress bar, not the part where tags are fetched?

src/sess.rs
```diff
@@ -968,34 +993,76 @@ impl<'io, 'sess: 'io, 'ctx: 'sess> SessionIo<'sess, 'ctx> {
...
-                    .spawn_with(move |c| c.arg("fetch").arg("--all").arg("--tags").arg("--prune"))
+                    .spawn_with(
+                        move |c| {
+                            c.arg("fetch")
+                                .arg("--all")
+                                .arg("--tags")
+                                .arg("--prune")
+                                .arg("--progress")
+                        },
+                        None,
```
micprog: Do we add the progress bar here?

src/error.rs
fischeti: Warnings are handled with `miette` now which also has a `Severity` trait. I removed it from here since it is was previously only used in the warnln! macro, but this does not exist anymore

log.txt
fischeti: whoops that slipped in, I removed it

src/error.rs
fischeti: good point, I will change the alignment of all stage messages in this way again. Should be cleaner

src/git.rs
fischeti: Yes, to my knowledge, this covers the main part of the clone resp. fetch. The problem is a bit that if you would track progress also for the second fetch, you would end up with two completion messages as well, which is a bit confusing.

I will actually just try it out to add progress bars here as well.

src/sess.rs
fischeti: to be honest, I am not sure why there is an additional fetch here. Here we are in the `ToCheckout`  stage, so my assumption is that it was already cloned, and we only need to checkout now.

src/error.rs
fischeti: Addressed in https://github.com/pulp-platform/bender/pull/239/commits/208fec0cafb0eb5b029f3bb668509abd27ed9a82

src/cmd/vendor.rs
fischeti: Cleaned up in https://github.com/pulp-platform/bender/pull/239/commits/670e436174013059ae9d506513a80e4bd32a3821

src/sess.rs
fischeti: I added a progress handler here as well, although I wasn't able to observe it. I think it is only called if the git database is outdated, which is a bit hard to reproduce.

fischeti:
I should have addressed all the comments now. Additionally, I added the following:

* https://github.com/pulp-platform/bender/pull/239/commits/0949dda9aa197e84ef9fae93929d4f64ac982315: All colored messages (e.g. stages, warnings etc.,) will now only be collored in environments that allow colors i.e. if `NO_COLOR` is not set in the environment. For instance `bender command 2> command.log` will not contain any ANSI codes anymore.
* https://github.com/pulp-platform/bender/pull/239/commits/2f884e5667298c96e7f4177ade32284ab4ea5e4f: I aligned the formatting of `update` command during manual version resolving to the ones introduced with progress bars and the diagnostics (This is also why the CLI regression test fails at the moment, since I removed the backticks when rendering paths.)
* https://github.com/pulp-platform/bender/pull/239/commits/bd380f4ea84f3e17feb9cac7d8a227d6f31dfb06: I changed to style of versions (cyan now) to differentiate it from the style of packages (white bold). I am not fully satisfied with the look tbh, so we can also revert this. Or if you have a better idea let me know.
* https://github.com/pulp-platform/bender/pull/239/commits/bb71e6238d9e87f9c7cbb7be7480756cd3e0417c: Fixes an issue where `stageln` messages interfered with the progress bars. This was very rare but I noticed it a few times when symlinks were created for dependencies, while cloning was in progress. Now `stageln` always suspends the progress bars to fix this issue.

PR #240, opened 2026-01-05
Refactor CLI, group flags, add aliases and support for env variables

fischeti:
# CLI refactor

## `derive` feature

Since a while `clap` supports [declaritive structs](https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html) instead of a runtime builder, which is now the [recommended way](https://docs.rs/clap/latest/clap/_faq/index.html#when-should-i-use-the-builder-vs-derive-apis) to build CLIs. It is easier to read, and modify and the resulting implementation is very clean in my opinion. Especially when it comes to subcommands which are represented as enums in Rust.

## CLI organisation

I reorganized all the CLI subcommands and flags, in a way that should still be backwards compatible:

* Script targets  i.e. `bender script <target>` are now implemented as nested subcommands of `script`. This change is mainly reflected when calling the help page of `script`.
* Some of the script targets where previously defined as global but were not used by all targets and had to be checked inside `script.rs`. Examples are `--vlog-arg`, `--vcom-arg` and `--only-*`. Now, they are moved/duplicated into each target that uses them which automatically takes care of these checks.
* The `--only-*` flags are grouped, meaning `clap` will automatically return an error if multiple one of them are specified at the same time. The same goes for `packages --graph`, that does not allow `--version` and/or `--targets`
* Some flags have been renamed i.e. `--vlog-arg` was used by all targets for general verilog arguments even though the underlying executable has a different name. Now, the flag is called `--vlogan-arg` for VCS, but backward compatibility is ensured with aliases to `--vlog-arg`.
* The help pages are now colored and the flags are now grouped into global options and subcommand specific options (+ subsubcommand options for `bender script <target> --help`).
* Several aliases were introduced, which do not show up on help pages but are compatible. Those mainly consist of the plural/singular version of the flag e.g. `--vlog-args`/`--vlog-arg` both work now or `parents`/`parent` as well.

## Environment variables

Global flags (e.g. git throttle, local, dir etc., ) can now also be provided as environment variables. The help pages also reflect this. The precedence is flag > environment variable > default.

## TODOs:

- [x] Fix plugins (or deprecate them?)
- [x] Remove comments after review (of previous CLI command builders)
- [x] Evaluate further optimizations (e.g. `bender script <target>` could be implemented as subsubcommands with enums instead of strings if possible)
- [x] Remove redundant `$[arg(long = <field-but-dashed>, short = <first-letter-of-field>]`
- [x] Potentially ~deprecate~ modify previous `--vcom-args` etc., in favor of tool-specific args ~(with graceful deprecation warning in the beginning)~ with aliases to keep backward compability
- [x] Consider using `env` feature of `clap` to set flags based on environment variables (and deprecate config?)
- [x] Fix Windows tests (or not?). Currently there is an inconsistency with new line delimiters. Fixed in https://github.com/pulp-platform/bender/pull/245
- [x] ~Decide how/whether to merge `cli_regression` test suite~ moved to https://github.com/pulp-platform/bender/pull/245
- [x] Change flag and add backward compatible aliases for flags that used `lower_case` i.e. `vendor` had some.
- [x] Address all inline TODOs
- [x] Review of consisteny between old CLI and new one ensure backward compability
- [x] Wait for https://github.com/pulp-platform/bender/pull/245 to be merged

### DIFF
--- a/src/cmd/script.rs
+++ b/src/cmd/script.rs
@@ -243,12 +243,12 @@ pub fn run(sess: &Session, args: &ScriptArgs) -> Result<()> {
             ScriptFormat::Vcs { .. } => vec!["vcs", "simulation"],
             ScriptFormat::Verilator { .. } => vec!["verilator", "synthesis"],
             ScriptFormat::Synopsys { .. } => vec!["synopsys", "synthesis"],
-            ScriptFormat::Formality { .. } => vec!["synopsys", "synthesis", "formality"],
+            ScriptFormat::Formality => vec!["synopsys", "synthesis", "formality"],
             ScriptFormat::Riviera { .. } => vec!["riviera", "simulation"],
-            ScriptFormat::Genus { .. } => vec!["genus", "synthesis"],
+            ScriptFormat::Genus => vec!["genus", "synthesis"],
             ScriptFormat::Vivado { .. } => concat(vivado_targets, &["synthesis"]),
             ScriptFormat::VivadoSim { .. } => concat(vivado_targets, &["simulation"]),
-            ScriptFormat::Precision { .. } => vec!["precision", "fpga", "synthesis"],
+            ScriptFormat::Precision => vec!["precision", "fpga", "synthesis"],
             ScriptFormat::Template { .. } => vec![],
             ScriptFormat::TemplateJson => vec![],
         }
@@ -257,12 +257,7 @@ pub fn run(sess: &Session, args: &ScriptArgs) -> Result<()> {
     };
 
     // Filter the sources by target.
-    let targets = TargetSet::new(
-        args.target
-            .iter()
-            .map(|s| s.as_str())
-            .chain(format_targets.into_iter()),
-    );
+    let targets = TargetSet::new(args.target.iter().map(|s| s.as_str()).chain(format_targets));
 
     if args.assume_rtl {
         srcs = srcs.assign_target("rtl".to_string());
@@ -344,7 +339,7 @@ pub fn run(sess: &Session, args: &ScriptArgs) -> Result<()> {
             tera_context.insert("vhdl_args", vhdl_args);
             include_str!("../script_fmt/synopsys_tcl.tera")
         }
-        ScriptFormat::Formality {} => include_str!("../script_fmt/formality_tcl.tera"),
+        ScriptFormat::Formality => include_str!("../script_fmt/formality_tcl.tera"),
         ScriptFormat::Riviera {
             vlog_args,
             vcom_args,
@@ -353,7 +348,7 @@ pub fn run(sess: &Session, args: &ScriptArgs) -> Result<()> {
             tera_context.insert("vcom_args", vcom_args);
             include_str!("../script_fmt/riviera_tcl.tera")
         }
-        ScriptFormat::Genus {} => include_str!("../script_fmt/genus_tcl.tera"),
+        ScriptFormat::Genus => include_str!("../script_fmt/genus_tcl.tera"),
         ScriptFormat::Vivado { no_simset, only } | ScriptFormat::VivadoSim { no_simset, only } => {
             only_args = only.clone();
             tera_context.insert("vivado_filesets", &{
@@ -365,12 +360,12 @@ pub fn run(sess: &Session, args: &ScriptArgs) -> Result<()> {
             });
             include_str!("../script_fmt/vivado_tcl.tera")
         }
-        ScriptFormat::Precision {} => include_str!("../script_fmt/precision_tcl.tera"),
+        ScriptFormat::Precision => include_str!("../script_fmt/precision_tcl.tera"),
         ScriptFormat::Template { template } => &std::fs::read_to_string(template)?,
-        ScriptFormat::TemplateJson => &JSON.to_string(),
+        ScriptFormat::TemplateJson => JSON,
     };
 
-    emit_template(sess, tera_context, &template_content, args, only_args, srcs)
+    emit_template(sess, tera_context, template_content, args, only_args, srcs)
 }
 
 /// Subdivide the source files in a group.


src/cli.rs
```diff
@@ -14,131 +13,129 @@ use std::fs::{canonicalize, metadata};
...
+    local: bool,
+
+    /// Sets the maximum number of concurrent git operations
+    #[arg(
+        long,
+        global = true,
+        help_heading = "Global Options",
+        env = "BENDER_GIT_THROTTLE"
+    )]
+    git_throttle: Option<usize>,
```
micprog, line 55: It seems to me the default value for this is no longer set, so the default value is no longer visible from the help menu. It may be beneficial to set this up, and then later on checking if the flag was passed or not when setting the value to be used.

src/cmd/vendor.rs
```diff
@@ -35,56 +35,52 @@ pub struct PatchLink {
...
+pub enum VendorSubcommand {
+    /// Display a diff of the local tree and the upstream tree with patches applied.
+    Diff {
+        /// Return error code 1 when a diff is encountered. (Optional) override the error message by providing a value.
+        // alias is for backward compatibility
+        #[arg(long, num_args(0..=1), alias="err_on_diff")]
+        err_on_diff: Option<String>,
+    },
+    /// (Re-)initialize the external dependencies.
+    Init {
```
micprog, line 67: The long_about seems to have gotten lost here.

src/cmd/update.rs
```diff
@@ -17,55 +17,34 @@ use crate::lockfile::*;
...
+
+    /// Update requested dependencies recursively, i.e., including their dependencies
+    #[arg(long, action = ArgAction::SetTrue, requires = "dep")]
+    pub recursive: bool,
 }
 
 /// Execute the `update` subcommand.
-pub fn setup(matches: &ArgMatches, suppress_warnings: &IndexSet<String>) -> Result<bool> {
-    let force_fetch = matches.get_flag("fetch");
-    if matches.get_flag("local") && matches.get_flag("fetch") && !suppress_warnings.contains("W14")
```
micprog, line 67: Looks like you missed checking for the local flag?

src/cmd/snapshot.rs
```diff
@@ -15,53 +15,34 @@ use crate::config::{Dependency, Locked, LockedSource};
...
+    /// Working directory to snapshot dependencies from
+    #[arg(long, default_value = "working_dir")]
+    pub working_dir: String,
+
+    /// Do not skip dependencies that are dirty
+    #[arg(long, action = ArgAction::SetTrue)]
+    pub no_skip: bool,
+
+    /// Checkout the dependencies snapshotted into the lockfile
+    #[arg(long, short = 'c', action = ArgAction::SetTrue)]
```
micprog: do you need to specify `= 'c'` here?

src/cmd/packages.rs
```diff
@@ -5,61 +5,38 @@
...
+/// Information about the dependency graph
+#[derive(Args, Debug)]
+#[command(alias = "package")]
+pub struct PackagesArgs {
+    /// Print the dependencies for each package
+    #[arg(short, long, action = ArgAction::SetTrue, conflicts_with_all = ["version", "targets"])]
+    pub graph: bool,
+
+    /// Do not group packages by topological rank
+    #[arg(short, long, action = ArgAction::SetTrue)]
```
micprog: long_help looks to have gotten lost

src/cmd/packages.rs
```diff
@@ -5,61 +5,38 @@
...
+    pub flat: bool,
+
+    /// Print the version of each package
+    #[arg(long, action = ArgAction::SetTrue)]
+    pub version: bool,
```
micprog, line 39: long_help looks to have gotten lost

src/cmd/parents.rs
```diff
@@ -15,35 +15,29 @@ use crate::error::*;
...
+#[derive(Args, Debug)]
+#[command(alias = "parent")]
+pub struct ParentsArgs {
+    /// Package name to get the parents for
+    #[arg(num_args(1))]
+    pub name: String,
+
+    /// Print the passed targets to the dependency
+    #[arg(long, num_args(0))]
+    pub targets: bool,
```
micprog: I think we should align the style of these arguments to the ones in other script formats.

micprog (changes requested):
Thanks for the additions! I really like the new format and what it enables in code cleanliness, as well as the predefined environment variables for certain flags. I haven't gone through the script.rs file yet, but here is some initial feedback.

src/cmd/snapshot.rs
fischeti: No not necessary. I removed it.

src/cmd/packages.rs
fischeti: You are right, I added it back

src/cmd/packages.rs
fischeti: You are right, I added it back

src/cmd/vendor.rs
fischeti: Yes, I added it back 👍

src/cmd/update.rs
fischeti: Whoops, I missed this good catch. I think its because its part of the global flags and not `UpdateArgs`

src/cmd/parents.rs
fischeti: You are right. While thinking about it `action = ArgAction::SetTrue` is completely redundant for `bool`. I removed it everywhere now.

src/cli.rs
fischeti: The problem with `Cli::parse()` compared to the previous `matches` is that there is no way of knowing where the value came from i.e. whether it was the default or the one specified by the flag/env. It would be possible to extract this again with matches, but it is a bit ugly:

```rs
#[arg(
    long,
    global = true,
    help_heading = "Global Options",
    env = "BENDER_GIT_THROTTLE",
    default_value = "4" 
)]
git_throttle: Option<usize>,

...

let matches = Cli::command().get_matches();
let cli = Cli::from_arg_matches(&matches)?;

// Determine git throttle. 
let git_throttle = if matches.value_source("git-throttle") == Some(clap::parser::ValueSource::DefaultValue) {
    config.git_throttle.or(cli.git_throttle).unwrap_or(4)
} else {
    cli.git_throttle.unwrap_or(4)
};
```

because you still need to set the default value in the code itself. The default in the argument derive it is just there to show up in the help message.

Instead, I just added it now in the help message, but let me know if you think this is better.

micprog: I think this makes sense, thanks!

micprog (approved):
LGTM, thanks!

PR #242, opened 2026-01-06
Remove unused `common-path` dependency + add CI check

fischeti:
Adds a workflow to the CI to check for unused dependency based on the `cargo-machete` plugin. It is said to be fast but not super precise since it is based on regex and does not actually build the project. Nevertheless, it already caught the unused `common-path` dependency.

The alternative would be `cargo-udeps` which should be more precise since it actually builds the project, but when trying it out, it did not compile and I didn't proceed it further.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -95,7 +95,6 @@ dependencies = [
  "blake2",
  "clap",
  "clap_complete",
- "common-path",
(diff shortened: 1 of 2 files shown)


micprog (approved):
LGTM :+1:

PR #244, opened 2026-01-09
Add metadata information to script output

mrogenmoserTT:


src/cmd/script.rs
```diff
@@ -535,7 +542,7 @@ fn emit_template(
 
     let mut all_defines = IndexMap::new();
     let mut all_incdirs = vec![];
-    let mut all_files = vec![];
+    let mut all_files = IndexSet::new();
```
fischeti, line 521: Makes sense to me, I was just wondering whether duplicate files caused any problems before

PR #246, opened 2026-01-11
Consolidate warnings, suppression and deduplication

fischeti:
## Warning collection

This PR collects the current warnings system into a monolithic `diagnostic::Warnings` enum, which has the following benefits:
* Make it much easier to keep warning codes coherent which are also used to suppress warnings
* Allows for easier reuse of warnings without duplicating the warning message
* Consistent formatting of Warning messages
* Additional help messages to guide the user to solution to the warning.

## Suppression handling improvements

Additionally, the current warning and error suppression system has been improved a bit. A global resp. static `Diagnostics` handler now takes care of suppression. It is initialized in the beginning with the user defined suppressed warnings. This cleans up the code since `suppress_warnings` does not need to be passed around anymore, and whether a warning is suppressed is checked automatically behind the scenes when a Warning is emitted. Further, the `Diagnostic` handler keeps track of emitted warnings and deduplicates them to avoid redundant warnings. It not only differentiates between the type of warning but also it's content (e.g. `SomeWarningAbougPkg` for `pkg1`  is treaded as different than `SomeWarningAboutPkg` for `pkg2`).

## Warning formatting

The new formatting of warnings and help messages is shown below:

```
warning: This is a warning without a code and without a help message.

warning[W03]: This is a warning with code "W03" and without help message
 
warning: This is a warning without a code, but a help message
 ╰─› help: This is the help message
 
warning[W02]: This is a warning with code "W02" and a help message
 ╰─› help: This is the help message
 
warning[W06]: This is a warning with code "W06" and multiple help messages (if separated by "\n")
 ├─› help: This is the first message
 ╰─› help: This is the second message
 ```
 
 (in reality the warnings are also colored properly)
 
 ## New crates adoption
 
Two new crates were adopted to reduce boilerplate code and derive warning/help messages, error codes and severity from procedural macros:

```rs
enum Warning {
    #[error("This is the warning message containing {some_data}")]
    #[diagnostic(code(W42), severity(warning), help("Fix something in {some_data}"))]
    SomeWarning { some_data: String }
}
```

### `miette`

This crate uses the `#[diagnostic]` macro to implement traits for the specified enumeration variants that allows it to call `help()`, `code()` and `severity()` on an enum variant. `miette` can do a lot more and is mostly used for nice annotations of source code. In the case of Bender this is less relevant since it is not a compiler that reads source code. It could in theory be used to annotate wrong `Bender.yml` files, but this is a bit overkill at the moment, and our current YAML parser does not return positional information of fields in the source.

### `thiserror`

The second crate uses the `#[error]` macro to derive the `Display` trait for the warning message. At the moment, we are not using `thiserror` to its full extend. It also implements automatic `Error` conversion and chaining, which will be implemented in a separate PR to make the review easier.

## TODO:

- [ ] Some warnings currently have the same warning code. This is not a problem for `miette`, but suppression is impacted potentially.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -128,8 +128,6 @@ pub fn main() -> Result<()> {
         })
         .collect();
 
-    let warn_config_loaded = !suppressed_warnings.contains("W02");
-
     // Initialize warning and error handling with the suppression arguments.
     Diagnostics::init(suppressed_warnings);
 
@@ -173,10 +171,7 @@ pub fn main() -> Result<()> {
     debugln!("main: {:#?}", manifest);
 
     // Gather and parse the tool configuration.
-    let config = load_config(
-        &root_dir,
-        matches!(cli.command, Commands::Update(_)) && warn_config_loaded,
-    )?;
+    let config = load_config(&root_dir, matches!(cli.command, Commands::Update(_)))?;
     debugln!("main: {:#?}", config);
 
     // Determine git throttle. The precedence is: CLI argument, env variable, config file, default (4).
--- a/src/diagnostic.rs
+++ b/src/diagnostic.rs
@@ -179,36 +179,28 @@ pub enum Warnings {
     #[diagnostic(code(W05))]
     NoFilesForGlobPattern { path: String },
 
-    // TODO(fischeti): Why are there two W06 variants?
     #[error("Dependency {} in checkout_dir {} is not a git repository. Setting as path dependency.", fmt_pkg!(.0), fmt_path!(.1.display()))]
-    #[diagnostic(
-        code(W06),
-        help("Use `bender clone` to work on git dependencies.\nRun `bender update --ignore-checkout-dir` to overwrite this at your own risk.")
-    )]
+    #[diagnostic(code(W06), help("Use `bender clone` to work on git dependencies.\nRun `bender update --ignore-checkout-dir` to overwrite this at your own risk."))]
     NotAGitDependency(String, PathBuf),
 
-    // TODO(fischeti): Why are there two W06 variants?
     #[error("Dependency {} in checkout_dir {} is not in a clean state. Setting as path dependency.", fmt_pkg!(.0), fmt_path!(.1.display()))]
     #[diagnostic(code(W06), help("Use `bender clone` to work on git dependencies.\nRun `bender update --ignore-checkout-dir` to overwrite this at your own risk."))]
     DirtyGitDependency(String, PathBuf),
 
-    // TODO(fischeti): This is part of an error, not a warning. Should be converted to an Error.
-    #[error("SSH key might be missing.")]
+    // TODO(fischeti): This is part of an error, not a warning. Could be converted to an Error in the future.
+    #[error("Failed to initialize git database.")]
     #[diagnostic(
         code(W07),
-        help("Please ensure your public ssh key is added to the git server.")
+        help("Please ensure the url is correct and you have access to the repository. {}",
+        if *is_ssh {
+            "\nEnsure your SSH keys are set up correctly."
+        } else {
+            ""
+        })
     )]
-    SshKeyMaybeMissing,
+    GitInitFailed { is_ssh: bool },
 
-    // TODO(fischeti): This is part of an error, not a warning. Should be converted to an Error.
-    #[error("SSH key might be missing.")]
-    #[diagnostic(
-        code(W07),
-        help("Please ensure the url is correct and you have access to the repository.")
-    )]
-    UrlMaybeIncorrect,
-
-    // TODO(fischeti): This is part of an error, not a warning. Should be converted to an Error.
+    // TODO(fischeti): This is part of an error, not a warning. Could be converted to an Error in the future.
     #[error("Revision {} not found in repository {}.", fmt_version!(.0), fmt_pkg!(.1))]
     #[diagnostic(
         code(W08),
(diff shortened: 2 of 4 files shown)


src/diagnostic.rs
```diff
@@ -0,0 +1,503 @@
...
+    CheckoutDirDirty(String, PathBuf),
+
+    #[error("Workspace checkout directory set and remote url doesn't match, not updating {} at {}.", fmt_pkg!(.0), fmt_path!(.1.display()))]
+    #[diagnostic(
+        code(W19),
+        help("Run `bender checkout --force` to overwrite the dependency at your own risk.")
+    )]
+    CheckoutDirUrlMismatch(String, PathBuf),
+
+    // TODO(fischeti): Should this be an error instead of a warning?
```
micprog: This is used in the resolver. Basically, the resolver attempts to match requirements with a dependency, but errors if there are no matches. This will error if the source (git URL) in the requirements matches with the currently tested dependency, but this may not always be the case. In cases where the source does not match, it may well be that the requirements cannot match but do for the required source, such as when certain commits or newer versions are only available in one of the two sources. As such, this mismatch is reported as a warning, only erroring when the requirement is unfulfillable. So, a warning is OK for this. It might make sense to update the code with this information.

src/diagnostic.rs
```diff
@@ -0,0 +1,503 @@
...
+    )]
+    UrlMaybeIncorrect,
+
+    // TODO(fischeti): This is part of an error, not a warning. Should be converted to an Error.
+    #[error("Revision {} not found in repository {}.", fmt_version!(.0), fmt_pkg!(.1))]
+    #[diagnostic(
+        code(W08),
+        help("Check that the revision exists in the remote repository or run `bender update`.")
+    )]
+    RevisionNotFound(String, String),
```
micprog: Can we convert this to a single warning, with a flag passed that enables the ssh key information? Ultimately we will want to put this into the help information of an error, but a warning is what we have for now...

src/diagnostic.rs
```diff
@@ -0,0 +1,503 @@
...
+    #[diagnostic(
+        code(W06),
+        help("Use `bender clone` to work on git dependencies.\nRun `bender update --ignore-checkout-dir` to overwrite this at your own risk.")
+    )]
+    NotAGitDependency(String, PathBuf),
+
+    // TODO(fischeti): Why are there two W06 variants?
+    #[error("Dependency {} in checkout_dir {} is not in a clean state. Setting as path dependency.", fmt_pkg!(.0), fmt_path!(.1.display()))]
+    #[diagnostic(code(W06), help("Use `bender clone` to work on git dependencies.\nRun `bender update --ignore-checkout-dir` to overwrite this at your own risk."))]
+    DirtyGitDependency(String, PathBuf),
```
micprog: I believe this was due to the warnings being closely related or an error on my part. Happy to adjust if needed, but IMO we can also keep as-is. Why is the formatting different for the two?

src/diagnostic.rs
```diff
@@ -0,0 +1,503 @@
...
+    #[error("File {} doesn't exist.", fmt_path!(path.display()))]
+    #[diagnostic(code(W31))]
+    FileMissing { path: PathBuf },
+
+    #[error("Path {} for dependency {} does not exist.", fmt_path!(path.display()), fmt_pkg!(pkg))]
+    #[diagnostic(code(W32))]
+    DepPathMissing { pkg: String, path: PathBuf },
+}
+
+#[cfg(test)]
```
micprog, line 324: Can we move these tests to a separate file in the tests directory?

src/cli.rs
```diff
@@ -125,9 +128,9 @@ pub fn main() -> Result<()> {
         })
         .collect();
 
-    if suppressed_warnings.contains("all") || suppressed_warnings.contains("Wall") {
-        suppressed_warnings.extend((1..24).map(|i| format!("W{:02}", i)));
-    }
+    let warn_config_loaded = !suppressed_warnings.contains("W02");
```
micprog: Is this still needed with the updated warning system, where these warning outputs are suppressed the diagnostics?

micprog (commented):
I added some minor things, but overall LGTM! I really like the updated output and the cleaner handling of suppressing warnings!

There are some warnings that should actually be help information for an error, which are currently noted as TODOs. Thanks for doing this, I think it's best addressed in an upcoming PR.

src/diagnostic.rs
fischeti: I see makes sense, I just wanted to double check

src/diagnostic.rs
fischeti: Yes good idea. It is now controlled by a `is_ssh` field. Turns out you can put arbitrary code in the derive macro for the help message.

src/diagnostic.rs
fischeti: Fine by me to keep it as is, just wanted to double check. The automattic formatting for derive macros is a bit random sometimes. I aligned the formatting now.

src/diagnostic.rs
fischeti: We can, this is the structure suggested by the [Rust book](https://doc.rust-lang.org/book/ch11-03-test-organization.html), but I don't have a strong opinion on that.

src/cli.rs
fischeti: No, good point. I simplified it now.

src/diagnostic.rs
micprog: Ok, then let's leave it as-is.

PR #247, opened 2026-01-11
Add package-specific targets and target removal from the command line

micprog:


src/src.rs
```diff
@@ -116,33 +116,90 @@ impl<'ctx> SourceGroup<'ctx> {
...
-        let chained_targets = TargetSet::new(targets.iter().chain(self.passed_targets.iter()));
+        let local_targets = TargetSet::new(
+            targets
+                .iter()
+                .filter_map(|trgt| {
+                    if !trgt.contains(':') {
+                        Some(trgt.clone())
+                    } else {
+                        match self.package {
+                            None => Some(trgt.clone()),
```
fischeti, line 127: This can only happen in the root package right? i.e. you could have a target `my_root:some_target` which would qualify for that branch.

src/src.rs
```diff
@@ -116,33 +116,90 @@ impl<'ctx> SourceGroup<'ctx> {
...
-                    .filter_targets(all_targets, use_passed)
+                    .filter_targets(targets, use_passed)
                     .map(|g| SourceFile::Group(Box::new(g))),
                 ref other => Some(other.clone()),
             })
             .collect();
+        let mut target_defs = all_targets
+            .into_iter()
+            .filter_map(|t| match t.contains(':') {
+                true => None,
```
fischeti: What is the idea here? If I understand it correctly `all_targets` are now stripped of the dependency, so this should never happen?

README.md
```diff
@@ -327,6 +327,10 @@ Additionally, we suggest to use the following targets to identify source code an
 - `rtl` for synthesizable RTL code
 - `gate` for gate-level netlists
 
+Do not use `:` in your custom targets, as this is used to separate targets to apply to individual packages.
+
+Do not start the target name with `-`, as this is used to remove target application.
```
fischeti, line 332: Potentially, it would make sense to simply forbid this and throw an error. I don't think that would be very restrictive.

micprog: Added errors if this is used in target specifications.

src/src.rs
micprog: Yes, I'll remove this filtering!

src/src.rs
micprog: I believe the root package will still populate the package field, so this should never actually be used.

PR #248, opened 2026-01-11
Add default trait to SourceGroup, TargetSpec, TargetSet

micprog:


fischeti (approved):
Love it😍 The only change I see is that `independent` defaults to `false` instead of `true` as before. But I don't think this should be an issue since `files` is just an empty list.

Issue #251, opened 2026-01-19
[bug] Fails to remove symlink on windows


sigmuha:
When using the `bender clone` command on Windows, it fails to remove the "old" checked out dependency's symlink:
```
error: Failed to remove symlink at path "XX". Access is denied. (os error 5)
```

As I understand it, symlinks on windows need to be removed with `std::fs::remove_dir(path)` and not `std::fs::remove_file(path)`, which can be seen in `src/cli.rs` 

https://github.com/pulp-platform/bender/blob/30ec37c758835a33e190fdb4c425a43848cf6eab/src/cli.rs#L266

and `src/cmd/clone.rs` 

https://github.com/pulp-platform/bender/blob/30ec37c758835a33e190fdb4c425a43848cf6eab/src/cmd/clone.rs#L275

PR #252, opened 2026-01-19
Added conditional compilation for symlink removal in cli and clone

sigmuha:
Removal of symlinks is now handled differently for Unix and Windows:
- Unix: Keeps `std::fs::remove_file()`
- Windows: Uses `std::fs::remove_dir()`instead. 
See #251 for issue

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -323,23 +323,23 @@ pub fn main() -> Result<()> {
     }
 }
 
-#[cfg(target_family = "unix")]
-fn symlink_dir(p: &Path, q: &Path) -> Result<()> {
+#[cfg(unix)]
+pub fn symlink_dir(p: &Path, q: &Path) -> Result<()> {
     Ok(std::os::unix::fs::symlink(p, q)?)
 }
 
-#[cfg(target_os = "windows")]
-fn symlink_dir(p: &Path, q: &Path) -> Result<()> {
+#[cfg(windows)]
+pub fn symlink_dir(p: &Path, q: &Path) -> Result<()> {
     Ok(std::os::windows::fs::symlink_dir(p, q)?)
 }
 
-#[cfg(target_family = "unix")]
-fn remove_symlink_dir(path: &Path) -> Result<()> {
+#[cfg(unix)]
+pub fn remove_symlink_dir(path: &Path) -> Result<()> {
     Ok(std::fs::remove_file(path)?)
 }
 
-#[cfg(target_os = "windows")]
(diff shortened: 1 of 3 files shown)


src/cmd/clone.rs
```diff
@@ -329,6 +329,18 @@ pub fn symlink_dir(p: &Path, q: &Path) -> Result<()> {
...
+pub fn remove_symlink_dir(path: &Path) -> Result<()> {
+    Ok(std::fs::remove_file(path)?)
+}
+
+/// Remove a directory symlink.
+#[cfg(windows)]
+pub fn remove_symlink_dir(path: &Path) -> Result<()> {
+    Ok(std::fs::remove_dir(path)?)
+}
+
```
micprog: Can you refer to the functions in the cli file for this? You should be able to include the following at the top and then you should be able to remove these declarations.

```
use crate::cli::remove_symlink_dir;
```

src/cli.rs
```diff
@@ -342,6 +342,16 @@ fn symlink_dir(p: &Path, q: &Path) -> Result<()> {
...
+#[cfg(target_family = "unix")]
+fn remove_symlink_dir(path: &Path) -> Result<()> {
+    Ok(std::fs::remove_file(path)?)
+}
+
+#[cfg(target_os = "windows")]
+fn remove_symlink_dir(path: &Path) -> Result<()> {
+    Ok(std::fs::remove_dir(path)?)
+}
+
```
micprog: For consistency, can we adjust the cfg to the following and add the `pub` for accessibility from other files?
```suggestion
#[cfg(unix)]
pub fn remove_symlink_dir(path: &Path) -> Result<()> {
    Ok(std::fs::remove_file(path)?)
}

#[cfg(windows)]
pub fn remove_symlink_dir(path: &Path) -> Result<()> {
    Ok(std::fs::remove_dir(path)?)
}

```

src/cmd/clone.rs
sigmuha: Will do! Should the creation functions also be removed for consistency? 
https://github.com/pulp-platform/bender/blob/ca6e3337d682a16829456166ce4028758abebc8f/src/cmd/clone.rs#L321-L330

micprog: Yes, that would be great! The goal is to reduce duplicate code and simplify what is there :+1:

src/cli.rs
sigmuha: Should the symlink creation functions also use change their cfg?

PR #254, opened 2026-01-19
Migrate to Rust 2024 Edition

fischeti:
Bumps the Rust edition from 2018 to 2024, and modernizes the project structure in the following way:

* A new `lib.rs` file collects all the modules (previously done in `main.rs`), which is the new entry point of bender. This technically means that bender can now also be used as a library crate and not only a binary crate.
* `main.rs` is moved to `bin/bender.rs` and calls the bender library now (with the `bender::` prefix)
* All `extern crate` are removed (not required anymore since the 2018 edition).
* `cmd/mod.rs` is replaced with `cmd.rs`, which collects all modules in `cmd`

Furthermore, new Rust editions typically change some formatting rules and some new clippy warnings surfaced (e.g. unused functions), which were fixed.

### DIFF
--- a/src/sess.rs
+++ b/src/sess.rs
@@ -649,89 +649,87 @@ impl<'io, 'sess: 'io, 'ctx: 'sess> SessionIo<'sess, 'ctx> {
             }
             None => {
                 debugln!("sess: git_versions get new");
-                let dep_refs = git.clone().list_refs().await;
-                let dep_revs = git.clone().list_revs().await;
-                let dep_refs_and_revs = dep_refs.and_then(|refs| -> Result<_> {
-                    if refs.is_empty() {
-                        Ok((refs, vec![]))
+                let dep_revs = git.clone().list_revs();
+                let dep_refs = git.clone().list_refs().await?;
+                let (refs, revs) = {
+                    if dep_refs.is_empty() {
+                        (dep_refs, vec![])
                     } else {
-                        dep_revs.map(move |revs| (refs, revs))
+                        (dep_refs, dep_revs.await?)
                     }
-                });
-                dep_refs_and_revs.map(move |(refs, revs)| {
-                    let refs: Vec<_> = refs
-                        .into_iter()
-                        .map(|(a, b)| (self.sess.intern_string(a), self.sess.intern_string(b)))
-                        .collect();
-                    let revs: Vec<_> = revs
-                        .into_iter()
-                        .map(|s| self.sess.intern_string(s))
-                        .collect();
-                    debugln!("sess: refs {:?}", refs);
-                    let (tags, branches) = {
-                        // Create a lookup table for the revisions. This will be used to
-                        // only accept refs that point to actual revisions.
-                        let rev_ids: IndexSet<&str> = revs.iter().copied().collect();
+                };
+                let refs: Vec<_> = refs
+                    .into_iter()
+                    .map(|(a, b)| (self.sess.intern_string(a), self.sess.intern_string(b)))
+                    .collect();
+                let revs: Vec<_> = revs
+                    .into_iter()
+                    .map(|s| self.sess.intern_string(s))
+                    .collect();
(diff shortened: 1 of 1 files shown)


src/bin/bender.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2017 ETH Zurich
+// Fabian Schuiki <fschuiki@iis.ee.ethz.ch>
+
+fn main() {
+    if let Err(e) = bender::cli::main() {
+        eprintln!("{}", e);
+        std::process::exit(1);
```
micprog: Is this part needed?

src/cmd.rs
```diff
@@ -5,8 +5,6 @@
 //!
 //! This module implements the subcommands of the command line tool.
 
-#![deny(missing_docs)]
```
micprog, line 8: Do we need to remove this? Documentation is good to have...

src/resolver.rs
```diff
@@ -1169,21 +1154,12 @@ struct DependencyReference<'ctx> {
     id: DependencyRef,
     /// The available versions of the dependency.
     versions: DependencyVersions<'ctx>,
-    /// The currently picked version.
-    pick: Option<usize>,
-    /// The available version options. These are indices into `versions`.
-    options: Option<IndexSet<usize>>,
 }
```
micprog, line 1177: Do I understand correctly these fields are not used anywhere?

src/sess.rs
```diff
@@ -653,7 +654,7 @@ impl<'io, 'sess: 'io, 'ctx: 'sess> SessionIo<'sess, 'ctx> {
                         dep_revs.map(move |revs| (refs, revs))
                     }
                 });
-                dep_refs_and_revs.and_then(move |(refs, revs)| {
+                dep_refs_and_revs.map(move |(refs, revs)| {
```
micprog: I'm not entirely sure this is the correct way to handle this... There may be a simpler way.

src/sess.rs
```diff
@@ -825,14 +826,13 @@ impl<'io, 'sess: 'io, 'ctx: 'sess> SessionIo<'sess, 'ctx> {
                     update_list,
                 )
                 .await
-                .and_then(move |path| {
+                .inspect(move |&path| {
```
micprog: I'm not sure this is the correct way to handle this. If I'm not mistaken, this is a remnant from the original futures implementation, that was updated with async/await, but can probably be simplified now.

src/sess.rs
```diff
@@ -1446,14 +1446,13 @@ impl<'io, 'sess: 'io, 'ctx: 'sess> SessionIo<'sess, 'ctx> {
                     Ok(None)
                 }
             })
-            .and_then(move |manifest| {
+            .inspect(move |&manifest| {
```
micprog: See above comments.

src/bin/bender.rs
fischeti: You still need to return a a non-zero exit code somehow. You could also return a result instead, which will actually do the same thing:

```rust
fn main() -> Result<(), Box<dyn std::error::Error>> {
    bender::cli::main()?;
    Ok(())
}
```

However, the error will be debug formatted which is a bit uglz

src/cmd.rs
fischeti: Actually no, I will revert this

src/resolver.rs
fischeti: Yes, exactly. clippy returns a warning in the new edition

src/bin/bender.rs
micprog: Ok, I see the point! The current setup drops the preceding "error:" in red, if I'm not mistaken, so some adjustment is still needed... Is this where using anyhow or thiserror could be useful?

micprog: I pushed a fix to return to the previous behavior.

PR #259, opened 2026-01-20
Add support for git repos with lfs

### LINKED ISSUE #258: Bender fails with lfs enabled repository
Bender failed to clone repositories containing LFS objects. This occurred because Bender clones from a local bare mirror (DB repo) which does not contain the LFS objects (since it's a bare clone). The default git clone behavior attempts to smudge LFS files immediately, failing when objects are missing.

I set up two repos for a minimal test case:

https://github.com/persimmonai/bender-lfs-debug-top
https://github.com/persimmonai/bender-lfs-debug

The bender-lfs-debug repo has a single github lfs tracked file. Bender's call to git fails with a message like

`Downloading random.bin (1.0 MB)
Error downloading object: random.bin (d2b02ab): Smudge error: Error downloading random.bin (d2b02ab60695736d28fe734d8432a9a416d92c14f3a610289d9c6ce4781478c0): error transferring "d2b02ab60695736d28fe734d8432a9a416d92c14f3a610289d9c6ce4781478c0": [0] remote missing object d2b02ab60695736d28fe734d8432a9a416d92c14f3a610289d9c6ce4781478c0`

uge:
Fixes #258 

Modified  src/sess.rs
 
Disable LFS smudging during the internal git clone / git checkout operation from the DB repo to the workspace checkout directory. This allows the checkout to succeed even with missing LFS objects. After checkout, configure lfs.url in the checkout directory to point to the original upstream repository URL.

Run git lfs pull to explicitly download the LFS objects from the upstream.

### DIFF
--- a/README.md
+++ b/README.md
@@ -405,6 +405,20 @@ overrides:
 # DEPRECATED: This will be removed at some point.
 plugins:
   additional-tools: { path: "/usr/local/additional-tools" }
+
+# Number of parallel git tasks. Optional.
+# Default: 4
+# The number of parallel git operations executed by bender can be adjusted to
+# manage performance and load on git servers. Can be overriden as a command
+# line argument.
+git_throttle: 2
+
+# Enable git lfs. Optional.
+# Default: true
+# Some git dependencies may use git-lfs for additional source files. As
+# fetching these files may not always be desired or requried, it can be
+# disabled. For multiple conflicting settings will use true.
+git_lfs: false
 ```
 
 [Relevant code](https://github.com/pulp-platform/bender/blob/master/src/config.rs)
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -483,6 +483,7 @@ fn load_config(from: &Path, warn_config_loaded: bool) -> Result<Config> {
         overrides: None,
         plugins: None,
         git_throttle: None,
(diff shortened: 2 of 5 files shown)


micprog (changes requested):
Thank you for this addition, I think it makes sense and is very useful in case lfs is enabled and needed!

Unfortunately, it will fail if `git-lfs` is not installed, which may be the case. Maybe we can disable the lfs functionality with the following:
- if there is a configuration in a bender configuration file (e.g., `Bender.local`) disabling this. This can be useful if the lfs files are known to not be required.
- if the repository doesn't require it. This way we save running the command and keep bender working for systems that don't currently have git-lfs installed.
- Maybe we can add a warning if it required but not supported (or disabled) instead of erroring out on the lfs comands completely.

My main concern is ensuring the existing systems and functionality do not break by adding this new feature. Happy to help out to make the required modifications.

I also see you added a file (`bender-lfs-debug-top`), I don't believe this is needed.

uge:
Added functionality per points 2 and 3 to check for system install of git-lfs and also a warning if not supported but required. Also should only run if the repo has lfs in .gitattribute. Not sure how to structure a solution for the first point - should it be for the local repository or for the entire dependency tree?

PR #260, opened 2026-01-21
Add default remotes to manifest

fischeti:
Allows for specifying (default) remotes for git dependencies, for a more concise specification. The new dependency specification more closely resembles `Cargo.toml` or `pyproject.toml`. The main difference is that those use predefined defaults (crates.io and PyPI), which are well established registries. In Bender this is not the case, so remotes have to be specified explicitly. This feature is also somewhat inspired from how `uv` handles multiple indexes (see [here](https://docs.astral.sh/uv/concepts/indexes/)).

```yaml
# Specify git remotes for dependencies. Optional.
remotes:
  pulp:
    url: "https://github.com/pulp-platform"
    default: true # Only required if multiple remotes are specified.

  # Additional non-default remotes (HTTP or SSH).
  openhw: "https://github.com/openhwgroup"
  internal: "git@gitlab.company.com:internal-repo"

# Other packages this package depends on. Optional.
dependencies:
  # Path dependency.
  axi: { path: "../axi" }

  # Git version dependency from default remote.
  apb: "0.2"

  # Git version dependency from non-default remote.
  fll: { version: "0.8", remote: "internal" }
  
  # Git version dependency with explicit git url.
  ara: { git: "https://github.com/john_doe/ara.git", version: "2" }
```

The feature is optional and also backward compatible, meaning specifying the full git url is still supported. Furthermore, this short notation is deliberatly not supported for git revision dependencies in order to highlight those dependencies and nudge people to create and use actual releases. 

It is also possible to set a different upstream name, in case the repository name is different from the package name:

```yaml
  floo_noc: {version: "0.7.0", upstream_name: "FlooNoC" }
```

However, this feature is not advertised in the README.md

_Note: The CLI regression tests are failing because I added more dependencies to the `Bender.yml`, which triggers the `audit` output._

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,9 @@ remotes:
 
   # Additional non-default remotes (HTTP or SSH).
   openhw: "https://github.com/openhwgroup"
-  internal: "git@gitlab.company.com:internal-repo"
+  # Template remote URL where `{}` is a placeholder for dependency name.
+  # If no placeholder is found, "<url>/{}.git" is used.
+  internal: "git@gitlab.company.com:internal-repo/{}/release"
 
 # Other packages this package depends on. Optional.
 dependencies:


src/config.rs
```diff
@@ -315,18 +315,27 @@ impl PrefixPaths for Workspace {
...
     type Output;
     /// The error type produced by validation.
     type Error;
     /// Validate self and convert into the non-partial version.
-    fn validate(
-        self,
-        package_name: &str,
-        pre_output: bool,
-    ) -> std::result::Result<Self::Output, Self::Error>;
+    fn validate(self, ctx: &ValidationContext) -> std::result::Result<Self::Output, Self::Error>;
```
micprog: I like the addition of a context for validation, as that simplifies some invocation! Can we use a different name than `ctx`? I don't want to confuse with the `'ctx` lifetime that is used in some cases.

src/config.rs
```diff
@@ -656,88 +717,130 @@ impl PrefixPaths for PartialDependency {
...
+            }),
+            (_, _, Some(_), Some(_), _) => Err(Error::new(format!(
+                "Dependency `{}` cannot specify both `version` and `rev` fields.",
+                ctx.package_name
+            ))),
+            (git, Some(_), rev, version, _)
+                if git.is_some() || rev.is_some() || version.is_some() =>
+            {
+                Err(Error::new(format!(
+                    "Path dependency `{}` cannot specify `git`, `rev`, or `version` fields.",
```
micprog: I think this error is not as easily identifiable:
```suggestion
                    "Dependency `{}` cannot simultaneously specify `path` with `git`, `rev`, or `version` fields.",
```

src/config.rs
```diff
@@ -656,88 +717,130 @@ impl PrefixPaths for PartialDependency {
...
+            }
+            (Some(_), _, None, None, _) => Err(Error::new(format!(
+                "Dependency `{}` with `git` must also specify `rev` or `version`.",
+                ctx.package_name
+            ))),
+            (None, None, None, None, _) => Err(Error::new(format!(
+                "Dependency `{}` must specify at least one of `path`, `git`, `rev`, or `version`.",
+                ctx.package_name
+            ))),
+            cfg => Err(Error::new(format!(
```
micprog, line 843: Would it be worth differentiating an error for given git and given remote?

src/config.rs
```diff
@@ -656,88 +717,130 @@ impl PrefixPaths for PartialDependency {
...
+            // ```yaml
+            // my_dep: "1.2.3"
+            // my_dep: {version: "1.2.3"}
+            // ```
+            (None, None, None, Some(version), None) => {
+                if let Some(default_remote) = ctx.default_remote {
+                    let git_name = self.upstream_name.unwrap_or(ctx.package_name.to_string());
+                    Ok(Dependency::GitVersion {
+                        target,
+                        url: format!("{}/{}.git", default_remote.url, git_name),
```
micprog: Could we consider adding a (very minimal) template for the URL? My thought is to keep the functionality you implemented, but maybe adding the option to insert a `{}` where the dependency name should be placed (thus supporting things like remote_url=`https://github.com/pulp-platform/{}.git` with dep `common_cells` turning into `https://github.com/pulp-platform/common_cells.git`). While this shouldn't be an issue for GitHub repos, I think supporting this functionality could be beneficial in general.

src/config.rs
fischeti: Ah yes of course, I didn't think about this.

src/config.rs
fischeti: yes, makes sense. I will add it 👍

src/config.rs
fischeti: Yes, I think that should not be too hard to implement.

fischeti: Done in https://github.com/pulp-platform/bender/pull/260/commits/cbbf9c531cf00d48d1a1dda9512d050a701d722f

src/config.rs
fischeti: I renamed it to `vctx` now

src/config.rs
fischeti: Done

src/config.rs
micprog: Thanks for the quick fix! Adding a quick note to the README explaining the template option would be great as well.

fischeti: done

PR #264, opened 2026-01-28
Add slang bindings and `pickle` command

fischeti:
This PR introduces first-class Slang integration to Bender via a new `bender-slang` crate and adds the new pickle command for preprocessing/rewrite/emit workflows on SystemVerilog sources.

## Slang bindings

Slang is vendored in as a submodule and built with CMake during crate build, then linked statically into `bender-slang` (see  [`build.rs`](https://github.com/pulp-platform/bender/blob/fischeti/slang-pickler/crates/bender-slang/build.rs))

Slang has a couple of other library dependencies (e.g. `mimalloc`, `fmt`) which are fetched by cmake and also statically linked to improve portability of the bender binary (at the expense of binary size). Those libraries are force-fetched to prevent using preinstalled system-libaries. I also investigated linking of `libc++` statically, but this depends on the OS:
* macOS: Only allows dynamic linking of libc++ apparently. But it never caused any problems so far.
* Linux: ~`build.rs` tries to link `libc++` statically, if no `libstdc++.a` exists, it falls back to dynamic linking. The idea was that this would also allow to run pre-compiled bender on older linux distributions without newer C++ features (Slang requires C++ 20)~ Link dynamically. We will ship binaries `manylinux_2_28` containers that should provide enough compatibility, the fallback is building from source.
* Windows: Linked dynamically, [ideally statically](https://github.com/rust-lang/rfcs/blob/master/text/1721-crt-static.md#future-work) but difficult to debug without access to a windows machine.

## Rust/C++ bridge design

The bindings between C++ and Rust are created with the [`cxx`](https://crates.io/crates/cxx) crate (from the one and only dtolnay). It allows to access native types from both sides (e.g. Rust's `String` and C++'s `std::string`), and `cxx` generates the necessary glue code underneath. It also statically checks at compile time for ABI mismatches etc.

You just need to create a bridge i.e. [`lib.rs`](https://github.com/pulp-platform/bender/blob/fischeti/slang-pickler/crates/bender-slang/src/lib.rs) on the Rust side and [slang_bridge.h](https://github.com/pulp-platform/bender/blob/fischeti/slang-pickler/crates/bender-slang/cpp/slang_bridge.h) on the C++ side that define the FFI. Additionally, `lib.rs` wraps the FFI to provide a more idiomatic Rust API.

## The `pickle` command

`pickle` is the first feature to use the Slang binding in the following way:

### Source collection

* Uses Bender manifest-driven source ordering/groups
* Supports additional external files via CLI positional args
* Supports additional include dirs (`-I`) and defines (`-D`)
* Non-Verilog files (e.g. VHDL/unknown) are skipped with warnings

### `--top` trimming

* Supports one or more top modules via `--top`
* Trimming granularity is syntax tree/file level, matching `slang --depfile-trim` behavior:
   * if a file defines multiple modules, it is retained/dropped as a whole
* Finer-grained declaration-level stripping is theoretically possible but would require deeper AST rewriting and is deferred for now.

### Renaming with `--prefix` and `--suffix`

* Prefix/suffix rename support for modules/packages/interfaces
* Explicit exclusions via `--exclude-rename`

### Print options

Re-emitting the SV code from the syntax trees can be done with the following (self-explanatory) options currently:

* `--expand-macros`
* `--strip-comments`
* `--squash-newlines`

Includes are always printed. Furthermore it is possible to print the Syntax trees instead of emitting SV code with `--ast-json`.

### Error/Warning behaviour

* Slang emits built-in diagnostics itself
* Slang parse failures are propagated into Bender as structured errors, and command exits with failure
* It would be possible to support permissive mode (e.g. downgrade to warnings), but not part of this PR

## Related Changes

File type classification to (e.g. `SourceType::Verilog`) was moved from `script.rs` to `sess.rs`.

## Open questions/TODOs
- [ ] `bender-slang` currently is a path dependency and likely needs to be published as well. Otherwise `cargo install -F slang` will not work, only from source installation is supported.
- [ ] Test and Release builds are WIP and in separate PRs (see below)

---
This is **part 1 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #282 
- <kbd>&nbsp;2&nbsp;</kbd> #272 
- <kbd>&nbsp;1&nbsp;</kbd> #264 👈

### DIFF
--- a/src/cmd/pickle.rs
+++ b/src/cmd/pickle.rs
@@ -17,7 +17,7 @@ use crate::diagnostic::Warnings;
 use crate::error::*;
 use crate::sess::{Session, SessionIo};
 use crate::src::{SourceFile, SourceGroup, SourceType};
-use crate::target::TargetSet;
+use crate::target::{TargetSet, TargetSpec};
 
 use bender_slang::{SlangPrintOpts, SlangSession, SyntaxTreeRewriter};
 
@@ -130,8 +130,8 @@ pub fn run(sess: &Session, args: PickleArgs) -> Result<()> {
         let include_dirs = args
             .include_dir
             .iter()
-            .map(|d| sess.intern_path(Path::new(d)))
-            .collect::<IndexSet<_>>();
+            .map(|d| (TargetSpec::Wildcard, sess.intern_path(Path::new(d))))
+            .collect();
         let defines = args
             .define
             .iter()
@@ -172,7 +172,7 @@ pub fn run(sess: &Session, args: PickleArgs) -> Result<()> {
             .include_dirs
             .iter()
             .chain(src_group.export_incdirs.values().flatten())
-            .map(|path| path.to_string_lossy().into_owned())
+            .map(|(_, path)| path.to_string_lossy().into_owned())
             .chain(args.include_dir.iter().cloned())
             .collect::<IndexSet<_>>()
             .into_iter()


README.md
```diff
@@ -518,6 +524,30 @@ Supported formats:
...
+This command is only available when Bender is built with Slang support (for example via `cargo install bender --all-features`).
+
+Useful options:
+- `--top <MODULE>`: Trim output to files reachable from one or more top modules.
+- `--prefix <PFX>` / `--suffix <SFX>`: Add a prefix and/or suffix to renamed symbols.
+- `--exclude-rename <NAME>`: Exclude specific symbols from renaming.
+- `--ast-json`: Emit AST JSON instead of source code.
+- `--expand-macros`, `--strip-comments`, `--squash-newlines`: Control output formatting.
+- `-I <DIR>`, `-D <DEFINE>`: Add extra include directories and preprocessor defines.
+
```
micprog, line 575: It might be beneficial to mention that similar options to script exist (target, no-deps, package, exclude, ...)

src/src.rs
```diff
@@ -338,16 +338,13 @@ impl<'ctx> SourceGroup<'ctx> {
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize)]
+#[serde(rename_all = "lowercase")]
 pub enum SourceType {
     /// A Verilog file.
     Verilog,
-    // /// A SystemVerilog file.
-    // SystemVerilog,
     /// A VHDL file.
     Vhdl,
-    /// Unknown file type
```
micprog, line 393: Why are we removing the unknown type?

src/cmd/script.rs
```diff
@@ -542,21 +534,17 @@ fn emit_template(
                             SourceFile::Group(_) => unreachable!(),
                         })
                         .collect(),
-                    file_type: match ty {
-                        SourceType::Verilog => "verilog".to_string(),
-                        SourceType::Vhdl => "vhdl".to_string(),
-                        SourceType::Unknown => "".to_string(),
-                    },
+                    file_type: Some(ty),
```
micprog, line 606: Does this affect the script outputs?

crates/bender-slang/README.md
```diff
@@ -0,0 +1,19 @@
+# bender-slang
+
+`bender-slang` provides the C++ bridge between `bender` and the vendored [Slang](https://github.com/MikePopoloski/slang) parser infrastructure.
```
micprog:
```suggestion
`bender-slang` provides the C++ bridge between `bender` and the [Slang](https://github.com/MikePopoloski/slang) parser infrastructure, included as a submodule.
```

crates/bender-slang/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+                "Failed to canonicalize include directory '{}': {}",
+                include, cause
+            ),
+        })?;
+        out.push(canonical.to_string_lossy().into_owned());
+    }
+    Ok(out)
+}
+
+#[cfg(not(windows))]
```
micprog: If I'm not mistaken, other parts of bender differentiate on `cfg(windows)` and `cfg(unix)`, might make sense to align

src/cmd/pickle.rs
```diff
@@ -0,0 +1,261 @@
...
+        session.reachable_trees(&args.top)?
+    };
+
+    let mut first_item = true;
+    for tree in trees {
+        let renamed_tree = tree.rename(
+            args.prefix.as_deref(),
+            args.suffix.as_deref(),
+            &args.exclude_rename,
+        );
```
micprog: Just to check - does this rename all modules/packages/interfaces, or only the ones actually declared within the file hierarchy? It may make sense to exclude renaming of modules/etc that are not declared in the scope to ensure they can be properly referenced to their assumed non-modified declarations.

micprog (approved):
This looks like a great feature, thanks for putting in the effort! Some small questions from my side below, generally more for my understanding. I would also appreciate if @phsauter or @paulsc96 could have a very quick look at the cpp files as I am not that familiar.

src/src.rs
fischeti: Mainly because `SourceType` is only ever used wrapped in an `Option` and I thought it would be better to represent `Some(Unknown)` simply as `None`.

src/cmd/script.rs
fischeti: No, I also adapted the serialization of `SourceType` to `#[serde(rename_all = "lowercase")]
` and the templates still interpret the type as the previous string

crates/bender-slang/src/lib.rs
fischeti: makes sense, I will align it.

src/cmd/pickle.rs
fischeti: This will rename all of them, including the ones from the dependencies. I would assume that in general, you would want to rename all modules/packages/interfaces in order to avoid any naming collisions later on, with other parts that use different versions. You can still exclude a module from being renamed if you want to reference it. How does morty handle this?

README.md
fischeti: done

src/cmd/pickle.rs
micprog: morty first [collects all declarations](https://github.com/pulp-platform/morty/blob/master/src/lib.rs#L56-L84) where these units are [renamed and registered for renaming](https://github.com/pulp-platform/morty/blob/master/src/lib.rs#L571-L579). morty only [renames instantiations where it knows it renamed the declarations](https://github.com/pulp-platform/morty/blob/master/src/lib.rs#L681-L691).

I think this makes a lot of sense, as you may not always know which modules are not declared inside the file list you're providing (although morty [does report which ones are used but not declared](https://github.com/pulp-platform/morty/blob/master/src/lib.rs#L418-L423)). This way, the pickled files, along with any others you may include separately (e.g., from a library or encrypted sources), still function correctly.

I think it makes more sense to ensure this default behavior, if desired an additional flag can be added to brute-force rename all instances irrespective of if their declarations are in the same file set.

fischeti: Good point, I reworked the rewriter to do it in two steps:
1. Collect all declared symbol names, and create a rename map
2. Rewrite only the symbols that are part of the rename map

This should be closer to what Morty currently does i.e. this allows for instance to exclude packages (with `--no-deps` or `--exclude <pkg>`), and later "link" it again with the original name.

micprog: Nice, thanks!

src/cmd/pickle.rs
```diff
@@ -0,0 +1,264 @@
...
+
+/// Pickle files
+#[derive(Args, Debug)]
+pub struct PickleArgs {
+    /// Additional source files to pickle, which are not part of the manifest.
+    files: Vec<String>,
+
+    /// The output file (defaults to stdout)
+    #[arg(short, long)]
+    output: Option<String>,
```
micprog, line 32: I like this option, it might make sense to add it to `script` in a future PR! Arguably it can be achieved with `>`, but why not add some flexibility.

src/cmd/pickle.rs
```diff
@@ -0,0 +1,264 @@
...
+    strip_comments: bool,
+
+    /// Squash newlines in the output
+    #[arg(long, help_heading = "Slang Options")]
+    squash_newlines: bool,
+
+    /// Dump the syntax trees as JSON instead of the source code
+    #[arg(long, help_heading = "Slang Options")]
+    ast_json: bool,
+}
```
micprog, line 89: It would be interesting to add an output for missing modules, either as a warning or a custom command. Future addition again I think, I know `morty` supported this, and we're checking for all internal declarations either way.

src/cmd/pickle.rs
```diff
@@ -0,0 +1,264 @@
...
+                    Some(path.to_string_lossy().into_owned())
+                }
+                // Vhdl or unknown file types are not supported by Slang, so we emit a warning and skip them.
+                SourceFile::File(path, _) => {
+                    Warnings::PickleNonVerilogFile(path.to_path_buf()).emit();
+                    None
+                }
+                // Groups should not exist at this point,
+                // as we have already flattened the sources.
+                _ => None,
```
micprog: Should this be unreachable?
```suggestion
                _ => unreachable!(),
```

crates/bender-slang/cpp/print.cpp
```diff
@@ -0,0 +1,39 @@
...
+// Prints the given syntax tree back to SystemVerilog source code,
+// with options to control the printing behavior
+rust::String print_tree(const shared_ptr<SyntaxTree> tree, SlangPrintOpts options) {
+    SyntaxPrinter printer(tree->sourceManager());
+
+    printer.setIncludeDirectives(options.include_directives);
+    printer.setExpandIncludes(true);
+    printer.setExpandMacros(options.expand_macros);
+    printer.setSquashNewlines(options.squash_newlines);
+    printer.setIncludeComments(options.include_comments);
```
micprog, line 24: Can the `--strip-comments` configuration be adapted to remove the entire line if it contains only a comment? Or is this why the `--squash-newlines` is necessary?

src/cmd/pickle.rs
fischeti: yes, why not

src/cmd/pickle.rs
fischeti: True, I added it because the emitted files can be super long, and `--output` allows to silence it (and maybe it's also a bit faster this way).

src/cmd/pickle.rs
fischeti: That is certainly possible. I will open some issue for future enhancements of pickle/bender-slang

fischeti: Tracked in https://github.com/pulp-platform/bender/issues/283

crates/bender-slang/cpp/print.cpp
fischeti: tbh, this is also a bit strange to me that `setIncludeComments(false)` only replaces comments with newlines, hence `--squash-newlines` is kind of always suggested when stripping comments.

PR #267, opened 2026-02-05
Add ability to override files

mrogenmoserTT:


src/config.rs
```diff
@@ -1191,11 +1200,17 @@ impl Validate for PartialSources {
                         .emit();
                     });
                 }
+                if override_files.is_some_and(|x| x)
```
fischeti: Maybe `override_files == Some(true)` is a bit clearer here.

src/cmd/script.rs
```diff
@@ -536,7 +586,26 @@ fn emit_template(
...
+                                            pkg,
+                                            p.to_string_lossy()
+                                        )),
+                                    },
+                                    None => FileEntry {
+                                        file: p.to_path_buf(),
+                                        comment: None,
+                                    },
+                                }
+                            }
```
fischeti, line 610: Why does this need to be done twice?

src/config.rs
mrogenmoserTT: Done

src/cmd/script.rs
mrogenmoserTT: I believe one applies to `all_files`, while the other applies to the source file groups. This could be updated in the future.

PR #272, opened 2026-02-17
Add testing infrastructure for slang

fischeti:
I added another testing repository in `tests/pickle` to test out multiple SystemVerilog options. The testing is done on multiple levels:

### Unit tests `basic.rs`

Just checks that parsing is sucessful or proper error is returned when broken SV code is provided.

### Refactoring tests `pickle.rs`

Tests renaming feature of `--prefix`/`--suffix` as well as file filtering with `--top`

### CLI regression `cli_regression.rs`

Existing test framework to check the emitted output against regression (to be enabled)

---
This is **part 2 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #282 
- <kbd>&nbsp;2&nbsp;</kbd> #272 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #264

### DIFF
--- a/tests/cli_regression.rs
+++ b/tests/cli_regression.rs
@@ -161,8 +161,4 @@ regression_tests! {
     packages:                       &["packages"],
     packages_graph:                 &["packages", "--graph"],
     packages_flat:                  &["packages", "--flat"],
-    // Enable once the golden binary is built with `slang` support.
-    // pickle_basic:                   &["pickle", "--target", "top"],
-    // pickle_top_trim:                &["pickle", "--target", "top", "--top", "top"],
-
 }


crates/bender-slang/cpp/slang_bridge.h
```diff
@@ -7,8 +7,9 @@
 #include "rust/cxx.h"
 #include "slang/diagnostics/DiagnosticEngine.h"
 #include "slang/diagnostics/TextDiagnosticClient.h"
-#include "slang/driver/Driver.h"
+#include "slang/parsing/Preprocessor.h"
 #include "slang/syntax/SyntaxTree.h"
+#include "slang/text/SourceManager.h"
```
micprog, line 12: I assume this is part of a cleanup, not actually changing the functionality?

crates/bender-slang/build.rs
```diff
@@ -64,6 +95,11 @@ fn main() {
     let dst = slang_lib.build();
     let lib_dir = dst.join("lib");
 
+    // Create a symlink for the generated include directory
+    if target_os == "linux" || target_os == "macos" {
+        refresh_include_symlink(&dst.join("include"));
+    }
+
```
micprog, line 100: Can we use the same `unix` guards here as in other places in the codebase for consistency?

crates/bender-slang/cpp/slang_bridge.h
fischeti: jep, when I added the clangd configuration, it told me that those headers are not used.

crates/bender-slang/build.rs
fischeti: yes, actually the if branch is redundant since the `refresh_include_symlink` is also defined for windows, where it is just a no op. I removed the if statements here.

PR #279, opened 2026-02-23
Modernize and improve error handling

fischeti:
The previous error implementation is outdated: both `cause` and `description` have been deprecated since a while in the standard library. This PR modernizes the error type and handling

## Adopting `miette` error type

The new global error type is now `miette::Report`, which has the following benefits:
* Error output follows the same style as warnings, and also reuses the same code since both warnings and errors are `Diagnostics`.
* Error messages can now contain help messages as well
* `miette::Report` has neat macros like `bail!`, `ensure!` and `miette!` (inspired by `anyhow`)
* The previous chaining is replaced with `wrap_err` (resp. `wrap_err_with` for lazy evaluation), which additionally has the advantage that the source is retained with the standard `source()` which allows to traverse the whole chain of errors and print it out.

## Suppressible errors

An additional `downgrade_if_suppressed()` function has been introduced to support downgrading errors to warnings. The suppressible errors are duplicated in an `Errors` enum with corresponding variants in `Warnings`. Then, the call site of the error can simply do:

```rust
Errors::FileMissing { ... }.downgrade_if_suppressed()?
```

## Restructuring

Additionally `error.rs` has been removed entirely, and the rest (e.g. `stageln!` and other macros) moved to the `diagnostic.rs`, where they also belong imo.


src/cmd/clean.rs
```diff
@@ -3,9 +3,11 @@
 use clap::Args;
 use std::path::Path;
 
+use miette::{Context as _, IntoDiagnostic as _};
 use std::fs;
 
-use crate::error::*;
+use crate::Result;
+use crate::infoln;
```
micprog, line 10: This is more of a question here, not something for this PR but future: would switching `infoln` to use `log::info` instead make sense?

src/cmd/fusesoc.rs
```diff
@@ -81,10 +85,10 @@ pub fn run_single(sess: &Session, args: &FusesocArgs) -> Result<()> {
     };
 
     if !file_str.contains(bender_generate_flag) {
-        Err(Error::new(format!(
+        Err(err!(
             "{}.core already exists, please delete to generate.",
             name
-        )))?
+        ))?
```
micprog: Is `bail!` more appropriate here?

src/cmd/pickle.rs
```diff
@@ -238,36 +241,35 @@ pub fn run(sess: &Session, args: PickleArgs) -> Result<()> {
 
     // Setup Output Writer, either to file or stdout
     let raw_writer: Box<dyn Write> = match &args.output {
-        Some(path) => Box::new(
-            File::create(path)
-                .map_err(|e| Error::new(format!("Cannot create output file: {}", e)))?,
-        ),
+        Some(path) => {
+            Box::new(File::create(path).map_err(|e| err!("Cannot create output file: {}", e))?)
```
micprog: Isn't this better handled with `wrap_err_with`?

src/cmd/script.rs
```diff
@@ -652,7 +655,8 @@ fn emit_template(
         "{}",
         Tera::default()
             .render_str(template, &tera_context)
-            .map_err(|e| { Error::chain("Failed to render template.", e) })?
+            .into_diagnostic()
+            .wrap_err("Failed to render template.")?
```
micprog, line 659: What's the difference in `wrap_err` and `wrap_err_with`?

src/cli.rs
```diff
@@ -417,15 +413,15 @@ fn find_package_root(from: &Path) -> Result<PathBuf> {
...
                     "No manifest (`Bender.yml` file) found. Stopped searching at filesystem boundary {:?}.",
                     path
-                )));
+                );
             }
         }
     }
 
-    Err(Error::new(
+    Err(err!(
```
micprog: I think this can also be a `bail!`?

src/git.rs
```diff
@@ -112,52 +153,41 @@ impl<'ctx> Git<'ctx> {
...
-        if let Some(mut stdout) = child.stdout.take() {
-            // We just read all of stdout.
-            if let Err(e) = stdout.read_to_end(&mut stdout_buffer).await {
-                return Err(Error::chain("Failed to read stdout", e));
-            }
-        }
+        let mut stdout = child
+            .stdout
+            .take()
+            .ok_or_else(|| err!("Failed to capture git stdout.",))?;
```
micprog: Is this not supposed to be something like a wrap_err?

src/git.rs
```diff
@@ -112,52 +153,41 @@ impl<'ctx> Git<'ctx> {
...
+            .read_to_end(&mut stdout_buffer)
+            .await
+            .into_diagnostic()?;
 
         // Wait for child process to finish
-        let status = child
-            .wait()
-            .await
-            .map_err(|e| Error::chain("Failed to wait on child", e))?;
+        let status = child.wait().await.into_diagnostic()?;
```
micprog: Should we add the error message extension?

src/git.rs
```diff
@@ -112,52 +153,41 @@ impl<'ctx> Git<'ctx> {
...
 
         // Join the stderr task to get the error log
-        let collected_stderr = stderr_handle
-            .await
-            .unwrap_or_else(|_| String::from("<internal error reading stderr>"));
+        let collected_stderr = stderr_handle.await.into_diagnostic()?;
```
micprog, line 173: Error message got lost - not sure if it is needed

src/git.rs
```diff
@@ -112,52 +153,41 @@ impl<'ctx> Git<'ctx> {
...
-                    format!(
-                        "Output of git command ({:?}) in directory {:?} is not valid UTF-8.",
-                        cmd, self.path
-                    ),
-                    cause,
-                )
-            })
+            String::from_utf8(stdout_buffer)
+                .into_diagnostic()
+                .wrap_err("Output of git command is not valid UTF-8.")
```
micprog: Can we add the git command back? Haven't encountered such an issue, but :shrug:

src/git.rs
```diff
@@ -232,7 +266,7 @@ impl<'ctx> Git<'ctx> {
             }))
         })
         .await
-        .map_err(|cause| Error::chain("Failed to join blocking task", cause))?
+        .into_diagnostic()?
```
micprog: Should we add a wrap_err?

src/diagnostic.rs
```diff
@@ -4,16 +4,62 @@
...
+/// Print debug information. Omitted in release builds.
+#[macro_export]
+#[cfg(debug_assertions)]
+macro_rules! debugln {
+    ($($arg:tt)*) => {
+        if $crate::diagnostic::ENABLE_DEBUG.load(std::sync::atomic::Ordering::Relaxed) {
+            $crate::diagnostic!($crate::fmt_with_style!("Debug", owo_colors::Style::new().blue().bold()); $($arg)*);
+        }
+    }
+}
```
micprog: Is this still necessary with #296

src/diagnostic.rs
```diff
@@ -106,6 +152,34 @@ impl Warnings {
...
+        let warning_code = self
+            .code()
+            .expect("All warning diagnostics must define a code")
+            .to_string();
+        let error_code = format!(
+            "E{}",
+            warning_code
+                .strip_prefix('W')
+                .expect("Warning code must start with `W`")
+        );
```
micprog: Are these messages returned at compile time or at runtime?

src/src.rs
```diff
@@ -547,19 +547,14 @@ impl<'ctx> Validate for SourceFile<'ctx> {
...
                 } else {
-                    Err(Error::new(format!(
-                        "[E31] File {} doesn't exist",
-                        env_path_buf.to_string_lossy()
-                    )))
+                    Warnings::FileMissing {
+                        path: env_path_buf.clone(),
+                    }
+                    .emit_or_error()?;
+                    Ok(SourceFile::File(path, ty))
```
micprog:
```suggestion
                if !exists {
                    Warnings::FileMissing {
                        path: env_path_buf.clone(),
                    }
                    .emit_or_error()?;
                }
                Ok(SourceFile::File(path, ty))
```

src/progress.rs
```diff
@@ -300,14 +300,8 @@ impl ProgressHandler {
...
             // Handle errors by finishing and clearing the target bar, then logging the error
-            GitProgress::Error(err_msg) => {
+            GitProgress::Error(_) => {
                 target_pb.finish_and_clear();
-                errorln!(
-                    "{} {}: {}",
-                    "Error during git operation of",
-                    fmt_pkg!(&self.name),
-                    err_msg
-                );
```
micprog, line 310: I assume this information is properly handled now?

src/resolver.rs
```diff
@@ -921,10 +926,10 @@ impl<'ctx> DependencyResolver<'ctx> {
                     let mut buffer = String::new();
                     io::stdin().read_line(&mut buffer).unwrap();
                     if buffer.starts_with('\n') {
-                        break Err(Error::new(format!(
+                        break Err(err!(
```
micprog: Does this work as a bail?

src/resolver.rs
```diff
@@ -988,7 +993,7 @@ impl<'ctx> DependencyResolver<'ctx> {
                 }
             }
         } else {
-            Err(Error::new(msg))
+            Err(err!(msg))
```
micprog: This looks like a bail.

src/sess.rs
```diff
@@ -1034,18 +1049,14 @@ impl<'io, 'sess: 'io, 'ctx: 'sess> SessionIo<'sess, 'ctx> {
                             },
                             None,
                         )
+                        .await
                         .map_err(|cause| {
```
micprog, line 1053: Can this be restructured using wrap_err?

src/sess.rs
```diff
@@ -734,14 +755,11 @@ impl<'io, 'sess: 'io, 'ctx: 'sess> SessionIo<'sess, 'ctx> {
                 })
                 .await
                 .map_err(move |cause| {
```
micprog, line 757: Can this be restructured unsing wrap_err?

src/sess.rs
```diff
@@ -688,14 +712,11 @@ impl<'io, 'sess: 'io, 'ctx: 'sess> SessionIo<'sess, 'ctx> {
...
```
micprog, line 714: Can this be restructured using wrap_err?

src/cmd/vendor.rs
```diff
@@ -107,31 +109,47 @@ pub fn run(sess: &Session, args: &VendorArgs) -> Result<()> {
...
-                    }).await?;
-                    let rev_hash = match vendor_package.upstream {
-                        config::Dependency::GitRevision { ref rev, .. } => Ok(rev),
-                        _ => Err(Error::new("Please ensure your vendor reference is a commit hash to avoid upstream changes impacting your checkout")),
-                    }?;
+                    git.clone()
+                        .spawn_with(|c| c.arg("clone").arg(url).arg("."), Some(pb))
+                        .await
+                        .map_err(move |cause| {
+                            Error::from(SessionErrors::GitDatabaseAccess(
```
micprog, line 116: Can this be implemented using wrap_err?

src/config.rs
```diff
@@ -644,21 +632,11 @@ impl Validate for PartialManifest {
...
-                                }
-                                .emit();
-                                None
-                            } else {
-                                Some(Err(Error::chain("[E30]", cause)))
-                            }
-                        }
+                .map(|(trgt, path)| {
+                    if !(vctx.pre_output || path.exists() && path.is_dir()) {
+                        Warnings::IncludeDirMissing(path.clone()).emit();
```
micprog: should this be emit_or_error instead of just emit? Not sure the functionality is preserved...

src/config.rs
```diff
@@ -1078,37 +1055,26 @@ impl Validate for PartialSources {
...
-                                }
-                                .emit();
-                                None
-                            } else {
-                                Some(Err(Error::chain("[E30]", cause)))
-                            }
+                            let warning = Warnings::IgnoredPath {
+                                cause: cause.to_string(),
+                            };
+                            warning.emit_or_error().err().map(Err)
```
micprog: Just curious how the `err().map(Err)` works, and if it preserves the Some/None behavior.

src/config.rs
```diff
@@ -1747,22 +1702,22 @@ impl Validate for PartialConfig {
...
-                None => return Err(Error::new("Git command or path to binary not configured")),
+                None => bail!("Git command or path to binary not configured"),
             },
             overrides: match self.overrides {
-                Some(d) => d.validate(vctx).map_err(|(key, cause)| {
-                    Error::chain(format!("In override `{}`:", key), cause)
-                })?,
+                Some(d) => d
+                    .validate(vctx)
+                    .map_err(|(key, cause)| cause.wrap_err(format!("In override `{key}`.")))?,
```
micprog, line 1714: Is the map_err -> wrap_err leveling necessary, or is a single wrap_err possible?

src/config.rs
```diff
@@ -1234,15 +1200,10 @@ impl Validate for PartialSources {
...
-                                            }
-                                            .emit();
-                                            None
-                                        } else {
-                                            Some(Err(Error::chain("[E30]", cause)))
-                                        }
+                                        let warning = Warnings::IgnoredPath {
+                                            cause: cause.to_string(),
+                                        };
+                                        warning.emit_or_error().err().map(Err)
```
micprog: as above

src/config.rs
```diff
@@ -1281,15 +1242,10 @@ impl Validate for PartialSources {
...
-                                }
-                                .emit();
-                                None
-                            } else {
-                                Some(Err(Error::chain("[E30]", cause)))
-                            }
+                            let warning = Warnings::IgnoredPath {
+                                cause: cause.to_string(),
+                            };
+                            warning.emit_or_error().err().map(Err)
```
micprog: as above

src/config.rs
```diff
@@ -1747,22 +1702,22 @@ impl Validate for PartialConfig {
...
+                Some(d) => d
+                    .validate(vctx)
+                    .map_err(|(key, cause)| cause.wrap_err(format!("In override `{key}`.")))?,
                 None => IndexMap::new(),
             },
             plugins: match self.plugins {
                 Some(d) => d
                     .validate(vctx)
-                    .map_err(|(key, cause)| Error::chain(format!("In plugin `{}`:", key), cause))?,
+                    .map_err(|(key, cause)| cause.wrap_err(format!("In plugin `{key}`.")))?,
```
micprog, line 1720: as above

micprog (commented):
Love the upgrade, there look to be some nice usability and styling improvements! I have some questions about some of the implementation, see comments.

src/cmd/script.rs
fischeti: `wrap_err_with ` accepts a closure, meaning it is lazily evaluated. I used it everywhere where you have a `format!` call i.e. it is a small optimization to only format the error message if an actual error is thrown. For static strings, I used `wrap_err` directly.

src/cli.rs
fischeti: you are right. I will change it.

src/git.rs
fischeti: `take()` returns an `Option` so `wrap_err` would not work here directly.

src/git.rs
fischeti: We could do that again. I think wrapping it mostly makes sense if the context of the underlying error message is unclear. And I am not sure whether "Failed to wait on child" makes it more obvious than what the `io::Error` would return, which is why I removed it.

src/git.rs
fischeti: This is actually already covered earlier when capturing `stderr`. If that does not succeed the `err!("Failed to capture git stderr",)` is thrown.

src/git.rs
fischeti: actually yes, I will add it back here

src/git.rs
fischeti: I will add it

src/diagnostic.rs
fischeti: no, you are right, I will remove it.

src/diagnostic.rs
fischeti: at runtime. I tried to construct them at compile time but it is hard because the code is not an actual field but can only be retrieved by the `.code()` method.

src/progress.rs
fischeti: this `errorln` call was actually redundant, because it does not actually return the error, but just prints out the stderr. This would happen anyway (later) when returning the actual error of the git operation.

src/resolver.rs
fischeti: probably, if `break` could be replaced with `return` then yes

src/resolver.rs
fischeti: jep, will change it.

src/sess.rs
fischeti: Not if we want to keep the help message of `LockedRevisionCheckout`. It is a bit ugly I agree, but the problem is that `wrap_err` does not accept an ad-hoc help message, which is why I created a specific error variant for that.

src/sess.rs
fischeti: same issue here

src/sess.rs
fischeti: same issue here

src/cmd/vendor.rs
fischeti: same issue here

src/config.rs
fischeti: you are right! I will change it.

src/config.rs
fischeti: yes, this is a bit weird, but is essentially short for:

```rust
match warning.emit_or_error() {
    Ok(()) => None,
    Err(e) => Some(Err(e)),
}
```

i.e. `emit_or_error()` returns a `Result`, `err()` converts from `Ok` into `None` resp. `Err(e)` into `Some(e)` and then `map(Err)` does `Some(Err(e))`. So the behaviour is preserved in my view. errors are not filtered out and the later collection will check if any of the elements is an error. The clanker says this would also be possible resp. more idiomatic:

```rust
.map(|entry| match entry.0.validate(vctx) {
    Ok(validated) => Ok(Some(validated)),
    Err(cause) => {
        let warning = Warnings::IgnoredPath {
            cause: cause.to_string(),
        };
        warning.emit_or_error()?;
        Ok(None)
    }
})
.collect::<Result<Vec<_>>>()?
.into_iter()
.flatten()
.collect()
```

since this decouples error handling and filtering

src/config.rs
fischeti: we need the `key` for context here. That's why I use `map_err`

src/cmd/clean.rs
fischeti: I think they should still be kept separate because `log::info` is potentially more verbose (timestamp, severity level) while `infoln` is more a user facing message with prettier formatting.

src/cmd/fusesoc.rs
fischeti: jep, I will change it.

src/cmd/pickle.rs
fischeti: yes, I will change it.

src/src.rs
fischeti: github tells me "Applying suggestions on deleted lines is currently not supported.", but I applied the suggestion in the most recent commit.

src/cmd/clean.rs
micprog: Ok :+1:

src/git.rs
micprog:
```suggestion
            .ok_or("Failed to capture git stdout.").into_diagnostic()?;
```

src/git.rs
micprog: I think a wrapper would help, even if the message is not 100% clear, to help trace it in the code for debug.

src/git.rs
micprog: Ok :+1:

src/diagnostic.rs
micprog: Let's change the default error warnings to have a code that starts with `E` to clarify that these are errors by default. It's a bit unfortunate we can't do these checks at compile-time, as a user will have a hard time understanding what's going wrong if this ever fires. OK otherwise :+1:

src/progress.rs
micprog: Ok :+1:

src/resolver.rs
micprog: I believe we return the error directly at line 951, so bail should be fine

src/config.rs
micprog: Ok, I understand it now, thanks!

src/config.rs
micprog: Ok :+1:

src/diagnostic.rs
fischeti: I tried and but it is ugly😅 The problem is the following:

If the default is to have "E" for the values, then for suppressed errors I need to construct an ad-hoc diagnostic `MietteDiagnostic` with the corresponding "W" code (severity is already a warning). However, I cannot call `emit()` on it, since it is not anymore a `Warnings` type...

fischeti: Potentially, it could make sense to create a new `Errors` enum and duplicate the downgradable errors in both `Errors` and `Warnings` enum. Then create a function, that explicitely downgrades an error variant to its warning variant. It's a bit more verbose, but it is at compile time

fischeti: also, you would explicitely have to define which errors are downgradable, which does actually make sense to me e.g.:

```rust
match error {
    Errors::Variant1 => Warnings::Variant1 
    Errors::Variant2 => Warnings::Variant2
    _ => unreachable!("This error is not downgradable")
}
```

or something like that

src/resolver.rs
fischeti: done

src/git.rs
fischeti: done

src/git.rs
fischeti: `into_diagnostic` does not seem to work for some reason, but I changed it to `ok_or`

src/diagnostic.rs
micprog: I think this is a good direction! Not all errors need to be in this enum, but downgradeable ones do. It makes things clear, and it checks at compile-time. Verbosity is an OK cost IMO.

fischeti: Did it in https://github.com/pulp-platform/bender/pull/279/commits/2f6d670e6fa1007e054f4354835c72b54d1ff9e1 and I think it does not look too bad. Also, I renamed `emit_or_error()` to `downgrade_if_suppressed()`, since the direction is reversed i.e. we are going from an Error to a Warning.

micprog: This makes things very clear now, thank you!

micprog (approved):
LGTM :+1:

PR #281, opened 2026-02-26
Add target filtering specifically for include directories

micprog:


src/config.rs
```diff
@@ -853,13 +868,63 @@ impl Validate for PartialDependency {
...
+}
+
 /// A partial group of source files.
 #[derive(Serialize, Deserialize, Debug, Default)]
 pub struct PartialSources {
     /// The targets for which the sources should be considered.
     pub target: Option<TargetSpec>,
     /// The directories to search for include files.
-    pub include_dirs: Option<Vec<String>>,
+    pub include_dirs: Option<Vec<StringOrStruct<PartialIncludeDir>>>,
```
fischeti, line 927: Does the source group also need to have a `PartialIncludeDir`? because those include directories already use the target from the SourceGroup.

micprog: A possible goal here is to support using different flavors of an include file, e.g., to define differing configuration variables. Yes, in the sourcegroup they already are behind targets, but this allows the files to have target-dependent includes without having to rewrite the file list multiple times.

PR #285, opened 2026-03-03
installer: replace `init` with version-aware router

fischeti:
Rename the old `init` script to `init-legacy` (unchanged) and introduce a thin version-aware router as the new init entry point.

- Versions >= `0.32.0` (first cargo-dist release) and `latest` are routed to the cargo-dist installer on GitHub Releases.
- Older versions are delegated to `init-legacy`, preserving full backward compatibility including the sh -s -- VERSION interface.

## TODO
- [x] Wait for https://github.com/pulp-platform/bender/pull/295 to be merged
- [x] Move scripts into `website` folder on `master`
- [x] Remove `website` branch


init
```diff
@@ -1,465 +1,42 @@
...
+#   curl --proto '=https' --tlsv1.2 https://pulp-platform.github.io/bender/init -sSf | sh -s -- 0.31.0
 
-set -u
+set -eu
 
-readonly GITHUB_API_BENDER_ROOT='https://api.github.com/repos/pulp-platform/bender'
-readonly GITHUB_API_HEADER='Accept: application/vnd.github.v3+json'
+REPO="pulp-platform/bender"
+# v0.32.0 is the first release with cargo-dist installers (all versions are 0.x.y)
+# TODO: revisit this comparison if bender ever reaches v1.x
```
micprog: Can you add a check and use the new form for anything 1.0.0 and beyond? Maybe defaulting to using the new flow and only using the legacy approach for <0.32 would make sense...

micprog (commented):
From what I understand, this will work with the new flow in #282. I think once that is merged and a new release is made (0.32), we immediately merge this to ensure the script remains properly functional with latest.

init
fischeti: Good call, I will change it.

micprog (approved):
LGTM :+1: Merging when we do the 0.32 release

fischeti:
Rebased this to `master` now so the `website` branch can be removed after this is merged

micprog:
@fischeti I updated the scripting to support the legacy flow and enable that as default, as well as enable global installs with a flag, including for legacy binaries. I haven't tested the script yet, but this well keep docs (e.g., worflows/ci) working. Let me know what you think!

fischeti:
LGTM now with https://github.com/pulp-platform/bender/pull/285/commits/015c0a12510e6b584c51f3225961183f28358645

PR #286, opened 2026-03-04
Clean up `sources` output

micprog:


src/src.rs
```diff
@@ -381,6 +381,109 @@ impl<'ctx> SourceGroup<'ctx> {
     }
 }
 
+/// A source file group.
+#[derive(Serialize, Clone, Debug, Default)]
+pub struct FilteredSourceGroup<'ctx> {
```
fischeti, line 386: A bit unfortunate that this requires another `*SourceGroup` type, but I also don't have a better solution.

PR #287, opened 2026-03-04
Add target filtering for defines

### LINKED ISSUE #40: Target-specific defines for all files
To avoid doubly listing files, would it be possible to declare defines for all files depending on the target? As an example, this would allow for tracing-specific signals and blocks to be instantiated without using bender's standard `TARGET_***` defines. Syntax could be similar to below.

```yaml
sources:
  - defines:
      FOO: { target: all(rtl, synth) }
```

micprog:
Fixes #40

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -111,7 +111,7 @@ where
             where
                 E: de::Error,
             {
-                self.visit_str(if value { "true" } else { "false" })
+                self.visit_str(&value.to_string())
             }
 
             fn visit_i64<E>(self, value: i64) -> std::result::Result<T, E>
--- a/README.md
+++ b/README.md
@@ -179,6 +179,8 @@ sources:
       EXCLUDE_MAGIC: ~
       # Define with a value.
       PREFIX_NAME: stuff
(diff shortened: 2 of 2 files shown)


src/util.rs
```diff
@@ -107,6 +107,34 @@ where
...
+                self.visit_str(&value.to_string())
+            }
+
+            fn visit_f64<E>(self, value: f64) -> std::result::Result<T, E>
+            where
+                E: de::Error,
+            {
+                self.visit_str(&value.to_string())
+            }
+
```
fischeti, line 137: Is it possible to write this as a generic function? E.g. something like:

```rust
            fn visit_bool<E>(self, value: T) -> std::result::Result<T, E>
            where
                T: ToString
                E: de::Error,
            {
                self.visit_str(&value.to_string())
            }       
```

micprog: I don't think we can have the visitor generic for all types, unfortunately. I can adapt the visit_bool function to match the pattern, but as serde already parses the bool it will remain a string version of the bool, not the original string.

PR #295, opened 2026-03-30
Bender Book

fischeti:
## TODOs
- [x] ~Change structure maybe? Put `book.toml` in root directory~
- [x] Figure out how to deploy to GH pages in a nice way (Github artifacts)
- [x] ~Align~ Mention in  https://github.com/pulp-platform/bender/pull/285, ~remove `website` branch~
- [x] Figure out what to do with `commands` section
- [x] Remove `sources` workflow?
- [x] Shorten README significantly and refer to the book
- [ ] Change pages deployment: Settings -> Pages -> Source: change from `website` to "Github Actions"


micprog:
@fischeti Does this already include the bender install script (and possibly updates from #285)? If we change the pointer from branch to pages, then the install script is no longer available (if I'm not mistaken). If possible, I'd like to ensure continuity on that front.

book/.gitignore
```diff
@@ -0,0 +1 @@
+book
```
micprog: nit: newline at end of file

book/src/workflows.md
```diff
@@ -0,0 +1,10 @@
+# Workflows
+
+This section provides practical guides on how to use Bender for common hardware development tasks throughout the project lifecycle.
+
+- **[Initialization](./workflow/init.md):** Setting up a new Bender package from scratch.
+- **[Adding and Updating Dependencies](./workflow/dependencies.md):** Managing external packages and ensuring reproducible builds.
+- **[Inspecting Source Files](./workflow/sources.md):** Exploring and filtering the source files in your project.
```
micprog: I don't think this part is around anymore.
```suggestion
```

fischeti:
No not yet. You are right, we should merge both of them at the same time. Let me rebase #285 onto this branch. Then we can merge it in here and push both book and new install scripts at once.

book/src/workflows.md
```diff
@@ -0,0 +1,10 @@
...
+- **[Generating Tool Scripts](./workflow/scripts.md):** Creating compilation and simulation scripts for various EDA tools.
+- **[Package Development](./workflow/package_dev.md):** Developing multiple packages simultaneously using the `clone` and `snapshot` flow.
+- **[Vendorizing External Code](./workflow/vendor.md):** Managing external repositories that don't natively support Bender.
```
micprog, line 9: Did we want to add the ci section here as well, or is this linked elsewhere?

micprog:
> No not yet. You are right, we should merge both of them at the same time. Let me rebase #285 onto this branch. Then we can merge it in here and push both book and new install scripts at once.

I see the website/init file in the diff now, while reviewing. It exists and looks to be placed in the correct spot, but it is still the old version. We will have to figure out how to get latest working correctly only once 0.32.0 is released, but I think we can manage.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,54 @@
+name: Deploy Documentation
```
micprog, line 1: Is this flow tested somewhere (e.g., on a temporary fork)? Just to ensure there are no issues here.

fischeti:
ah, right. I forgot that I also added it here already. This means this PR is independent from #285 and should not break anything regarding the installation script.

book/src/workflow/ci.md
```diff
@@ -0,0 +1,83 @@
...
+    with:
+      version: 0.31.0 # Optional: specify a version
+
+  - name: Checkout dependencies
+    run: bender checkout
+```
+
+## GitLab CI
+
+For GitLab CI, you can use the standard [installation script](../installation.md) to fetch the Bender binary at the start of your job.
```
micprog:
```suggestion
For GitLab CI, you can use the standard [installation script](../installation.md) to fetch the Bender binary at the start of your job if bender is not already installed in the system.
```

book/src/workflow/ci.md
```diff
@@ -0,0 +1,83 @@
...
+  stage: test
+  script:
+    - bender checkout
+    - bender script vsim > compile.tcl
+    - # Run your simulation tool here...
+```
+
+## Caching
+
+Since `Bender.lock` uniquely identifies the state of all dependencies, it is theoretically possible to cache the `.bender` directory to speed up pipelines. However, for most projects, the overhead of managing the cache (uploading/downloading) might outweigh the time saved by `bender checkout`, especially with fast network connections to Git remotes.
```
micprog: A potential improvement here is that the database directory can be shared across CI runs and even projects if the checkouts are properly partitioned. So, currently, if the toplevel `Bender.yml` specifies a `checkout_dir` (ensuring a checkout-specific dependency list), a `Bender.local` (e.g., in a parent directory, as it will get picked up there) can specify a common location for the database using `database: some/shared/location`. This drastically simplifies the fetches required, reduces the storage required, and, correspondingly, speeds up the CI. Some problems currently:
- If `checkout_dir` is not specified in the toplevel project's `Bender.yml`, checkouts may collide and cause issues. We can fix this within bender by separating the `Bender.local` configuration for `database_dir` and `checkout_dir`, but we will need to figure out what happens on `checkout_dir` collision with `Bender.yml`.
- Bender currently does not acquire a lock prior to git operations on such shared databases. Correspondingly, multiple concurrent runs may simultaneously act on this shared database and cause git errors that cause that particular run to fail. While this is unlikely, it is worthe mentioning if this information is added to the docs.

book/src/workflow/dependencies.md
```diff
@@ -0,0 +1,71 @@
+# Adding and Updating Dependencies
+
+As working with dependencies is one of bender's main features, there are a few commands to ensure functionality and assist with understanding the dependency structure.
+
+## New dependencies
+Once new dependencies are added to the manifest, bender needs to first be made aware of their existence. Otherwise, some commands will not work correctly and return an error. To update dependencies, run the following command:
```
micprog:
```suggestion
Once new dependencies are added to the manifest, bender needs to first be made aware of their existence. Otherwise, the internally used dependency tree will be incorrect or mismatched, and some commands will not work correctly, returning an error. To update dependencies, run the following command:
```

book/src/workflow/dependencies.md
```diff
@@ -0,0 +1,71 @@
...
+
+```sh
+bender update
+```
+
+In case other dependencies already exist and you do not want to re-resolve these, you can add the `--new-only` flag to the update command.
+
```
micprog, line 13: `Bender.lock` is referred to below, but not really introduced here. A quick note can fix that.

```suggestion

> **Note:** On update, bender creates or modifies the [`Bender.lock` lockfile](./lockfile.md), which keeps track of the currently selected dependency versions.

```

book/src/workflow/dependencies.md
```diff
@@ -0,0 +1,71 @@
...
+### Recursive Updates
+By default, updating a single dependency will not update its own child dependencies. If you want to update a package and all of its dependencies recursively, use the `-r/--recursive` flag:
+
+```sh
+bender update <PKG_NAME> --recursive
+```
+
+## Checking out Dependencies
+
+While `bender update` resolves versions and updates the lockfile, it does not necessarily download all the source code. To ensure all dependencies are locally available, use:
```
micprog: It may make sense to increase cross-linkages (e.g., lockfile here, and other spots).

book/src/workflow/dependencies.md
micprog:
```suggestion

> **Note:** On update, bender creates or modifies the [`Bender.lock` lockfile](./lockfile.md), which keeps track of the currently selected dependency versions.

> **Note:** Most bender commands will automatically run an update if no `Bender.lock` lockfile is found.

```

book/src/workflow/dependencies.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Finding Package Paths
+
+If you need to know where a specific dependency is stored (e.g., to point an external tool to it), use:
+
+```sh
+bender path <PKG_NAME>
+```
+
+This will output the absolute path to the package's root directory.
```
micprog: It might make sense to reference that this is the checkout directory, and I believe the command also checks out the specific dependency if it is not already checked out.

book/src/workflow/dependencies.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Checking Usage (`parents`)
+
+If you are wondering why a specific dependency was included or why a certain version was forced, you can use:
+
+```sh
+bender parents <PKG_NAME>
+```
+
+This will show you all packages in your tree that depend on `<PKG_NAME>` and what version constraints they have placed on it.
```
micprog, line 77: One big item missing here IMO is the `audit` command - this command can really help with checking the up-to-date-ness of dependencies as well as any potential problems.

book/src/workflow/dependencies.md
```diff
@@ -0,0 +1,71 @@
...
+
+```sh
+bender checkout
+```
+
+Bender will download the exact revisions specified in `Bender.lock`. This command is safe to run multiple times; it will only download missing packages or update those that have changed in the lockfile.
+
+> **Note:** Many other commands (like `sources` or `script`) will automatically trigger a checkout if they detect missing dependencies.
+
+## Inspecting the Dependency Tree
```
micprog: Can we link directly to the command's subfield in `commands.md`?

```suggestion
## Inspecting the Dependency Tree (`packages`)
```

book/src/workflow/dependencies.md
```diff
@@ -0,0 +1,71 @@
...
+
+To see the full list of dependencies and how they relate to each other, use the `packages` command:
+
+```sh
+bender packages
+```
+
+This prints a tree structure showing every package in your project, its resolved version, and its source. Use the `-f/--flat` flag if you just want a simple list of names and versions.
+
+## Finding Package Paths
```
micprog:
```suggestion
## Finding Package Paths (`path`)
```

book/src/bender_files.md
```diff
@@ -0,0 +1,31 @@
...
+| :--- | :--- | :--- | :--- |
+| **Role** | Manifest (Intent) | Lockfile (Reality) | Local Overrides |
+| **Main Content** | Dependencies & Version Ranges | Exact Git Revisions | Local Paths & Tool Config |
+| **Managed By** | User (Manual) | `bender update` (Auto) | User or `bender clone` |
+| **Version Control** | **Commit** | **Commit** | **Ignore** (`.gitignore`) |
+| **Shared?** | Yes, with everyone | Yes, for reproducibility | No, unique to your machine |
+
+## Summary of Roles
+
+### `Bender.yml` (The Manifest)
```
micprog: Let's link to `manifest.md`?

book/src/bender_files.md
```diff
@@ -0,0 +1,31 @@
...
+This is your **Intent**. It defines the requirements of your package. You use it to specify which other packages you need (e.g., "I need `axi` version `0.21.x`"). It is the only file required to define a Bender package.
+
+### `Bender.lock` (The Lockfile)
```
micprog: Let's link to `lockfile.md`?

book/src/bender_files.md
```diff
@@ -0,0 +1,31 @@
...
+This is the **Reality**. It is a machine-generated file that captures the exact state of your dependency tree. It records the specific Git commit hash for every dependency. By committing this file, you ensure that every developer and CI machine works with the exact same source code.
+
+### `Bender.local` (Local Overrides)
```
micprog: Let's link to `configuration.md`/`local.md`?

book/src/configuration.md
```diff
@@ -0,0 +1,62 @@
...
+The directory where Bender stores cloned and checked-out dependencies.
+- **Default:** `.bender` in the project root.
+- **Example:** `database: /var/cache/bender_dependencies`
+
+### `git`
+The command or path used to invoke Git.
+- **Default:** `git`
+- **Example:** `git: /usr/local/bin/git-wrapper.sh`
+
+### `git_throttle`
```
micprog, line 36: It may be worth mentioning that this (and below) can also be set (or overriden?) from envvar/command line arg

book/src/configuration.md
```diff
@@ -0,0 +1,62 @@
...
+Bender uses a cascading configuration system that allows you to set global defaults while still providing the flexibility to override settings for specific projects or local environments.
+
+## Configuration Files
+
+Bender looks for configuration files in the following order, merging them as they are encountered:
+
+1.  **System-wide:** `/etc/bender.yml`
+2.  **User-specific:** `$HOME/.config/bender.yml` (or the equivalent on your OS)
+3.  **Project-specific (checked in):** `.bender.yml` in your project root.
+4.  **Local Workspace (ignored):** `Bender.local` in your project root.
```
micprog: It may make sense to link to `local.md` or unify the contents?

book/src/dependencies.md
```diff
@@ -0,0 +1,88 @@
...
+Bender is designed to manage complex, hierarchical dependency trees for hardware projects. A dependency is any external package that provides its own `Bender.yml` manifest.
+
+## Dependency Types
+
+Dependencies are defined in the `dependencies` section of your `Bender.yml`.
+
+### Git Dependencies
+Git is the primary way to distribute Bender packages. You can specify them in two ways:
+
+> **Important:** The shorthand notation (e.g., `common_cells: "1.21.0"`) is only available if you have defined at least one [remote](#remotes) in your manifest. If no remote is specified, you must use the full `git` URL (see [revision](#revision-based))
```
micprog: I think this should be moved to after its introduction...

book/src/dependencies.md
```diff
@@ -0,0 +1,88 @@
...
+
+#### Version-based (Recommended)
+Bender uses [Semantic Versioning (SemVer)](https://semver.org/) to find the best compatible version. You can use standard [SemVer operators](https://docs.rs/semver/latest/semver/enum.Op.html) to specify version ranges:
```
micprog:
```suggestion
Bender uses [Semantic Versioning (SemVer)](https://semver.org/) to find the best compatible version. You can use [SemVer operators](https://docs.rs/semver/latest/semver/enum.Op.html) to specify version ranges:
```

book/src/dependencies.md
```diff
@@ -0,0 +1,88 @@
...
+
+```yaml
+dependencies:
+  common_cells: "1.21.0"
+  axi: { version: ">=0.23.0, <0.26.0" }
```
micprog, line 26: I think keeping a "standard" version example here (i.e., with git url) is important...

book/src/dependencies.md
```diff
@@ -0,0 +1,88 @@
...
+
+Dependencies can be conditionally included or configured using targets. For details on how to use target expressions or pass targets to dependencies, see the [Targets](./targets.md) documentation.
+
+## Git LFS Support
+
+Bender automatically detects if a dependency uses **Git Large File Storage (LFS)**. If `git-lfs` is installed on your system, Bender will automatically pull the required large files during the checkout process.
+
+## Submodules
+
+TODO
```
micprog: umm...

book/src/dependencies.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+### Path Dependencies
+Path dependencies point to a local directory. They are never versioned; Bender simply uses the code found at that location.
+```yaml
+dependencies:
+  my_local_ip: { path: "../local_ips/my_ip" }
+```
+
+## Remotes
```
micprog, line 45: Are we missing an explanation for a differing repo name from dep name for a given remote?

book/src/workflow/dependencies.md
```diff
@@ -0,0 +1,71 @@
...
+## New dependencies
+Once new dependencies are added to the manifest, bender needs to first be made aware of their existence. Otherwise, some commands will not work correctly and return an error. To update dependencies, run the following command:
+
+```sh
+bender update
+```
+
+In case other dependencies already exist and you do not want to re-resolve these, you can add the `--new-only` flag to the update command.
+
+## Updating dependencies
```
micprog, line 18: I think somewhere in this file (or in the dependencies.md), we need to mention what update does: It picks the highest compatible semver version and unique revision. If there are conflicts in the requirements and it cannot match all criteria, it asks the user. I didn't see this in dependencies.md or workflow/dependencies.md, but I think we need it somewhere to clarify what's going on internally.

book/src/index.md
```diff
@@ -0,0 +1,27 @@
...
+Bender helps you manage complex hardware IP hierarchies, ensuring that every member of your team and every CI runner is working with the exact same source code.
+
+## Key Features
+
+- **Hierarchical Dependency Management:** Resolve and manage transitive dependencies across multiple Git repositories or local paths.
+- **Reproducible Builds:** A precise lockfile mechanism (`Bender.lock`) ensures bit-identical design states across environments.
+- **HDL-Aware Source Collection:** Automatically manages file ordering, include directories, and preprocessor defines for SystemVerilog and VHDL.
+- **Target-Based Filtering:** Use powerful boolean expressions to include or exclude files based on your flow (simulation, synthesis, etc.).
+- **Local Development Workflow:** Easily modify dependencies in-place using the `clone` and `snapshot` flow without breaking official manifests.
+- **Tool Script Generation:** Generate compilation and simulation scripts for major EDA tools like QuestaSim, VCS, Vivado, Verilator, and more.
```
micprog, line 14: This seems like a bit much, but is accurate so :shrug:

```suggestion
## Key Features

- **Hierarchical Dependency Management:** Resolve and manage transitive dependencies across multiple Git repositories or local paths, allowing for IPs to be used across projects.
- **Reproducible Builds:** A precise lockfile mechanism (`Bender.lock`) ensures identical design states across environments.
- **HDL-Aware Source Collection:** Automatically manages file ordering, include directories, and preprocessor defines for SystemVerilog and VHDL.
- **Target-Based Filtering:** Use powerful boolean expressions to include or exclude files based on your flow (simulation, synthesis, etc.) or design configuration.
- **Local Development Workflow:** Easily modify dependencies in-place using the `clone` and `snapshot` flow without breaking official manifests.
- **Tool Script Generation:** Generate compilation and simulation scripts for major EDA tools like QuestaSim, VCS, Vivado, Verilator, and more.
```

book/src/installation.md
```diff
@@ -0,0 +1,64 @@
+# Installation
+
+Bender is a single standalone binary. You can either use our recommended shell installer, download a precompiled version, or build it from source.
+
+## Recommended: Shell Installer
+
+The fastest way to install Bender is using our shell script. It detects your operating system and architecture, downloads the latest release, and places it in your path (typically `~/.cargo/bin`).
```
micprog, line 7: Does the install script allow you to place the binary not where you're running the script? I don't remember... Also, for those who do not have cargo set up, the cargo/bin path makes little sense.

```suggestion
The fastest way to install Bender is using our shell script. It detects your operating system and architecture and downloads the latest release.
```

book/src/installation.md
```diff
@@ -0,0 +1,64 @@
...
+
+```sh
+cargo install --path .
+```
+
+## Linux Distributions
+
+Some Linux distributions provide Bender through their package managers:
+
+- **ArchLinux:** [Bender on the AUR](https://aur.archlinux.org/packages/bender)
```
micprog: brew?

book/src/local.md
```diff
@@ -0,0 +1,44 @@
+# Local Configuration (`Bender.local`)
+
+`Bender.local` is an optional, user-specific configuration file used to override the project's default settings. Its primary purpose is to allow local development of dependencies without modifying the shared `Bender.yml` or `Bender.lock`.
```
micprog: Might make sense to link or merge with configuration.md

book/src/local.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Overriding Dependencies
+
+The most common use for `Bender.local` is the `overrides` section. This forces Bender to use a specific version or a local path for a dependency, regardless of what the manifest requires.
+
```
micprog, line 10: It might be worth mentioning that these overrides are configured the same way dependencies are in the manifest dependencies section (described in dependencies.md afaik)(excluding the target stuff). It is also worth noting that these dependencies will only override existing configurations, not add new dependencies to any package.

book/src/local.md
```diff
@@ -0,0 +1,44 @@
...
+# Change the directory where dependencies are stored (default is .bender)
+database: my_deps_cache
+
+# Use a custom git binary or wrapper
+git: /usr/local/bin/git-wrapper.sh
+```
+
+## Best Practices
+
+- **Don't Commit It:** `Bender.local` should **rarely** be checked into version control. It contains paths and settings specific to your local machine. Always add it to your `.gitignore`.
```
micprog:
```suggestion
- **Don't Commit It:** `Bender.local` should **rarely** be checked into version control. It can contain paths and settings specific to your local machine. Always add it to your `.gitignore`.
```

book/src/manifest.md
```diff
@@ -0,0 +1,112 @@
...
+workspace:
+  # Create symlinks to checked-out dependencies.
+  package_links:
+    links/axi: axi
+    common: common_cells
+
+  # Directory where dependencies will be checked out. Optional.
+  # Once set, Bender performs the initial checkout and then leaves the directory
+  # untouched. Useful for chip packages that commit all dependencies into their
+  # own version control.
```
micprog: This behavior has changed...
```suggestion
  # Once set, Bender performs the initial checkout and then leaves the directory
  # untouched if any modification is detected. Useful for chip packages that commit
  # all dependencies into their own version control.
```

book/src/principles.md
```diff
@@ -0,0 +1,44 @@
+# Principles
+
+Bender was created to solve the challenges of managing large-scale hardware designs with complex, nested dependencies. It is built around four core principles that guide its development and usage.
+
+## 1. Modular and Opt-in
+Bender is designed to be a "pre-build" tool. It does not replace your EDA tools (synthesis, simulation, formal); it orchestrates them.
```
micprog:
```suggestion
Bender is designed to be a "pre-build" tool. It does not replace your EDA tools (synthesis, simulation, formal); it helps orchestrate them. 
```

book/src/principles.md
```diff
@@ -0,0 +1,44 @@
+# Principles
+
+Bender was created to solve the challenges of managing large-scale hardware designs with complex, nested dependencies. It is built around four core principles that guide its development and usage.
+
+## 1. Modular and Opt-in
+Bender is designed to be a "pre-build" tool. It does not replace your EDA tools (synthesis, simulation, formal); it orchestrates them. 
+- **Tool Agnostic:** Whether you use Vivado, Questa, VCS, or Verilator, Bender provides the necessary file lists and configurations.
```
micprog: These are only simulator tools, I would not specify an exhaustive list...

book/src/principles.md
```diff
@@ -0,0 +1,44 @@
+# Principles
```
micprog, line 1: This section seems a bit sloppy... I get the marketing pitch, but how different is the messaging here from the contents of index.md?

book/src/manifest.md
```diff
@@ -0,0 +1,112 @@
+# Manifest (`Bender.yml`)
+
+The package manifest describes the package, its metadata, its dependencies, and its source files. All paths in the manifest may be relative, in which case they are understood to be relative to the directory that contains the manifest. A manifest is required for each bender package.
```
micprog, line 3: It might make sense to mention here (or in a better location?) that generally for bender projects, the Bender.yml needs to be located at the project root, otherwise the git repo dependencies won't work properly. We could consider updating this in bender, but currently that's the behavior.

book/src/sources.md
```diff
@@ -0,0 +1,103 @@
...
+    defines:
+      USE_FAST_ALU: ~        # Define without a value
+      BIT_WIDTH: 64          # Define with a value
+    target: synthesis
+    files:
+      - src/rtl/alu.sv
+      - src/rtl/top.sv
+```
+
+- **include_dirs**: Paths added to the `+incdir+` flag during compilation.
```
micprog: It might be valuable to specify that it is intended for these include directories to only apply to the files in this (and child) sourcegroup.

book/src/targets.md
```diff
@@ -0,0 +1,172 @@
+# Targets
+
+Targets are the primary mechanism in Bender for managing project configurations. They allow you to conditionally include source files, include directories, and dependencies based on the current context (e.g., simulation vs. synthesis, or FPGA vs. ASIC).
```
micprog:
```suggestion
Targets are the primary mechanism in Bender for managing project configurations. They allow you to conditionally include source files, include directories, and dependencies based on the current context (e.g., simulation vs. synthesis, or FPGA vs. ASIC) or the desired configuration (e.g., with/without cache).
```

book/src/workflow/package_dev.md
```diff
@@ -0,0 +1,46 @@
...
+
+**Why use Snapshot?**
+The main benefit of a snapshot is portability. Because the lockfile is updated with the specific commit hashes, you can commit `Bender.lock` and share it with colleagues or run it in CI. The other environments will download the exact revisions you were working on from the Git remotes, without needing access to your local development paths.
+
+## Finalizing Changes
+
+Once your changes are stable and you are ready to "release" them:
+
+1.  **Tag the Dependency:** Push your changes to the remote repository and create a new version tag (e.g., `v1.2.2`).
+2.  **Update Manifest:** Update the version requirement in your top-level `Bender.yml` to include the new version.
```
micprog:
```suggestion
2.  **Update Manifest:** Update the version requirement in your `Bender.yml` to include the new version.
```

book/src/workflow/scripts.md
```diff
@@ -0,0 +1,86 @@
...
+
+The `script` command provides several flags to fine-tune the generated output:
+
+### Package Filtering
+- **`-p/--package <PKG>`**: Only include source files from the specified package (and its dependencies).
+- **`-n/--no-deps`**: Exclude all dependencies. This generates a script containing only the files from the current package or the packages explicitly listed with `-p`.
+- **`-e/--exclude <PKG>`**: Exclude a specific package from the generated script.
+
+### RTL Assumption
+- **`--assume-rtl`**: Automatically adds the `rtl` target to any source group that does not have an explicit target specification. This is a common shorthand for generating synthesis scripts without having to tag every RTL file.
```
micprog:
```suggestion
- **`--assume-rtl`**: Automatically adds the `rtl` target to any source group that does not have an explicit target specification. This is an optional shorthand for generating synthesis scripts without having to tag every RTL file.
```

book/src/workflow/scripts.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Relative Paths
+For generic file lists (`flist` or `flist-plus`), you can force Bender to use paths relative to the current directory using the `--relative-path` flag. This is useful for sharing file lists between different machines or environments.
+
+### Custom Templates
+Bender uses the [Tera](https://keats.github.io/tera/) templating engine for its scripts. If the built-in formats don't meet your needs, you can provide your own template file:
+
+```sh
+bender script template --template my_custom_format.tera > output.txt
+```
```
micprog, line 86: To assist in development, bender can dump the json it provides to the template using bender script template_json

book/src/workflow/vendor.md
```diff
@@ -0,0 +1,91 @@
+# Vendorizing External Code
+
+Bender's `vendor` command allows you to manage external dependencies that aren't natively packaged for Bender. It works by copying a subset of files from an upstream Git repository into a local directory within your project, allowing you to track changes and maintain local patches.
+
+This flow is heavily inspired by the `vendor.py` script used in the [OpenTitan](https://opentitan.org/) project.
```
micprog: Can we add a direct link to the script?

book/src/workflow/vendor.md
```diff
@@ -0,0 +1,91 @@
...
+
+### 1. Initialize (`init`)
+To download the upstream code and copy it into your `target_dir`:
+
+```sh
+bender vendor init
+```
+
+Bender clones the upstream repository, filters the files based on your rules, and copies them to your project. If a `patch_dir` is specified, any existing patches are applied automatically.
+
```
micprog, line 47: Can we add a note or some information here (or in a better spot) that the files listed in vendor are not automatically included as source files for the bender project! These need to be additionally specified in the source list in the manifest!

micprog (commented):
I think there are some points to address... I have not checked the README (I want to ensure all content previously covered within is properly covered in the book) and the commands (I want to ensure all are present), but figured I'd share the comments I already have.

book/src/installation.md
fischeti: Not anymore actually. The previous "installation" script only downloaded it to the current working directory but you would have to add it to the path manually.

In the cargo-dist configuration I currently use `CARGO_HOME` as the installation path. It doesn't even matter too much whether `CARGO_HOME` is already in the path or not i.e. if the user has a rustup installation. If it is not in the path, cargo-dist will adapt `.profile` to make sure that it is in the PATH next time the shell is (re)-started, which should also be prompted to the user during installation (see [here](https://axodotdev.github.io/cargo-dist/book/installers/shell.html#adding-things-to-path)).

This is breaking behaviour, and in theory we could also change the installation folder to the current working directory I guess. But in my opinion it does make much more sense to actually _install_ it somewhere in `$HOME` which is also aligned with other installation methods i.e. `cargo (b)install`

book/src/installation.md
fischeti: yes! will add it. Maybe with a mention that this is not maintained by us directly.

book/src/installation.md
fischeti: One option would be to allow fallback installation paths:

```toml
[dist]
install-path = ["$BENDER_INSTALL_DIR", "$HOME/.cargo/bin"]
```

I think this would for instance work with [mempool](https://github.com/pulp-platform/mempool/blob/3855df6da1893a5d18f34f3feaca5486d681738e/Makefile#L23-L24), although I am not a fan of installing bender this way i.e. per project instead of user.

.github/workflows/docs.yml
fischeti: will test it on my fork👍

book/.gitignore
fischeti: done

book/src/workflows.md
fischeti: yes, I linked it here as well now.

book/src/workflow/dependencies.md
fischeti: makes sense. I linked all `Bender.yml`, `Bender.lock` and `Bender.local` references to its section everywhere.

book/src/bender_files.md
fischeti: done

book/src/bender_files.md
fischeti: done

book/src/bender_files.md
fischeti: done

book/src/configuration.md
fischeti: makes sense. I restructured the entire configuration section to mention the three options: flag, env variable & configuration file, if exists.

book/src/configuration.md
fischeti: I linked it to `local.md` now. Or what do you mean exactly by unify the contents?

micprog: the configuration file and the Bender.local are essentially the same thing, so the contents should be more-or-less duplicate. I think the current form is reasonable with the link, though.

.github/workflows/docs.yml
fischeti: Required a small fix, but works now: https://fischeti.github.io/bender/

book/src/workflow/ci.md
micprog: done

book/src/workflow/dependencies.md
micprog: done

book/src/workflow/dependencies.md
micprog: done

book/src/installation.md
micprog: I think enabling backwards compatibility, even if it requires additional flags to be set, is a good thing. I agree that a system installation makes sense in general, but for a lot of environments it is actually preferrable to have an easy way to get a compatible binary (and then a sysadmin can put it where it's needed).

micprog (approved):
LGTM now :+1:

We can figure out how to handle the init script in #285

PR #297, opened 2026-04-02
script: filter out unreachable files

fischeti:
### script

Similar to the `pickle` command, the `script` command gains a new `--top` flag (can be specified multiple times), that will filter out all unreachable files from said top modules.

### `bender-slang`

Changes resp. fixes the behaviour for duplicate defined symbols. The previous behaviour was actually wrong (also in the `pickle` command) because it was first-one-wins. Now it is last-one-wins which is also what all EDA tools do afaik.

Further, it adds a slang diagnostic language whenever a module name already exists. The diagnostic is slang style and reuses slang infrastructure for this.

#### Integration tests

There are now integration tests for `script flist` output to check for existence of files (or the lack thereof). The pickle repo test suite is extended for that with an example of a duplicate module definition.

#### Real-world tests

Tested in cheshire for `vsim` and `vcs` target [here](https://github.com/pulp-platform/cheshire/tree/fischeti/bender-filter-ci-check), [pipeline](https://iis-git.ee.ethz.ch/github-mirror/cheshire/-/pipelines/109312) passes. Reduces file list from 670 to 444.

Also tested in FlooNoc for `vcs` and `synopsys` (elaboration), where it even discovered a duplicate file in the bender manifest🤓 

### TODOs

- [ ] Bump `bender-slang` once approved

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -148,7 +148,7 @@ dependencies = [
 
 [[package]]
 name = "bender-slang"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
  "cmake",
  "cxx",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ inherits = "release"
 lto = "thin"
 
 [dependencies]
-bender-slang = { version = "0.1.0", path = "crates/bender-slang", optional = true }
+bender-slang = { version = "0.1.1", path = "crates/bender-slang", optional = true }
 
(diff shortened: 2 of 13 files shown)


micprog:
One question - does this also filter out the include directories if they are not required? I think that would be useful when restricting the file list.

fischeti:
The include file filtering looks good to me. I would rebase this onto master, or should I wait for https://github.com/pulp-platform/bender/pull/309?

micprog (approved):
Thanks for this addition!

PR #299, opened 2026-04-16
feat: filter export_include_dirs of excluded packages by default

micprog:
When excluding dependencies via `--exclude`, their export include directories were still included in the output. Now they are filtered out by default. A new `--keep-excluded-incdirs` flag is available on `script`, `sources`, and `pickle` commands to preserve the old behavior.


fischeti (approved):
LGTM! I was only wondering whether the `--keep-excluded-dirs` flag is really necessary or whether this bloats the CLI.

PR #306, opened 2026-05-19
Add `bender kg` command and knowledge graph subsystem

aottavianoTT:
This PR introduces a knowledge graph (KG) pipeline that parses SystemVerilog designs via `bender-slang` into a queryable graph, exposed via `bender kg` CLI and an MCP server for AI assistant integration.

### Architecture

The pipeline is split across six new crates:

| Crate | Role |
|---|---|
| `bender-kg-models` | Shared IR types (`ModuleData`, `PortInfo`, `ParamInfo`, `InstantiationInfo`, …) used as the common data contract between crates |
| `bender-kg-extract` | Drives `bender-slang` to parse SV source files and emit structured IR |
| `bender-kg-store` | Persists the graph using Grafeo, augmented with an HNSW vector index for similarity lookups and BM25 for keyword search |
| `bender-kg-similarity` | Produces dense per-module embeddings via a Model2Vec/ONNX pipeline, powering structural similarity queries |
| `bender-kg-core` | Orchestrates the above into a typed `Engine` API; callers issue structured queries and get typed results back |
| `bender-kg-mcp` | Wraps `bender-kg-core` in a stdio MCP server so AI assistants can call all queries as MCP tools |

### Query surface

Once the KG is built (or incrementally updated), it supports:

- **Module search** — find modules by name or natural-language description
- **Hierarchy traversal** — walk instantiation trees up and down
- **Port / parameter / signal tracing** — trace connectivity across module boundaries
- **Structural similarity** — find modules that are structurally similar based on dense embeddings
- **Connectivity checks** — verify expected connections exist in the design

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,10 +11,47 @@ description = "A dependency management tool for hardware projects."
 readme = "README.md"
 license = "Apache-2.0 OR MIT"
 edition = "2024"
-rust-version = "1.87.0"
+# 1.91.1 matches Grafeo's MSRV.
+rust-version = "1.91.1"
 
 [workspace]
-members = ["crates/bender-slang"]
+members = [
+    "crates/bender-slang",
+    "crates/bender-kg-models",
+    "crates/bender-kg-extract",
+    "crates/bender-kg-store",
+    "crates/bender-kg-similarity",
+    "crates/bender-kg-core",
+    "crates/bender-kg-mcp",
+]
+
+[workspace.dependencies]
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+thiserror = "2.0.18"
+miette = "7.6.0"
+log = "0.4"
+sha2 = "0.10"
+clap = { version = "4.0", features = ["derive", "env"] }
+indexmap = { version = "2", features = ["serde"] }
+anyhow = "1"
+tokio = { version = "1.27", features = ["macros", "rt", "rt-multi-thread"] }
+# Embedded graph + vector + text store. The `embedded` profile bakes in
+# the LPG model + GQL parser + HNSW vectors + BM25 text index + algos +
+# parallel execution + single-file `.grafeo` storage; we add `cypher` for
+# our query strings and `wal` for crash-safe writes. We use a path dep
+# to the vendored Grafeo checkout next to the workspace.
+grafeo = { path = "../grafeo/crates/grafeo", default-features = false, features = ["embedded", "cypher", "wal"] }
+grafeo-common = { path = "../grafeo/crates/grafeo-common" }
+rmcp = { version = "1", features = ["server", "macros", "transport-io"] }
+schemars = "1"
+bender-slang = { path = "crates/bender-slang", version = "0.1.0" }
+bender-kg-models = { path = "crates/bender-kg-models", version = "0.1.0" }
+bender-kg-extract = { path = "crates/bender-kg-extract", version = "0.1.0" }
+bender-kg-store = { path = "crates/bender-kg-store", version = "0.1.0" }
+bender-kg-similarity = { path = "crates/bender-kg-similarity", version = "0.1.0" }
+bender-kg-core = { path = "crates/bender-kg-core", version = "0.1.0" }
+bender-kg-mcp = { path = "crates/bender-kg-mcp", version = "0.1.0" }
 
 [profile.dist]
 inherits = "release"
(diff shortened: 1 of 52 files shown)


Cargo.toml
```diff
@@ -63,4 +105,11 @@ pretty_assertions = "1.4"
...
 default = ["slang"]
-slang = ["dep:bender-slang"]
+slang = [
+    "dep:bender-slang",
+    "dep:bender-kg-core",
+    "dep:bender-kg-extract",
+    "dep:bender-kg-similarity",
+    "dep:bender-kg-mcp",
+    "dep:bender-kg-models",
+]
```
micprog, line 115: I would prefer if we could gate the new bender-kg behind an additional feature flag, disabled by default. This really looks like an AI-first feature, and I would argue we can keep it disabled for most use-cases.

README.md
```diff
@@ -587,6 +587,56 @@ bender pickle --top my_top --prefix p_ --suffix _s --exclude-rename my_top
 ```
 
 
+### `kg` --- Build and query the design knowledge graph
```
micprog, line 590: In a rebase, this will need to be moved to the book.

Cargo.toml
```diff
@@ -11,17 +11,59 @@ description = "A dependency management tool for hardware projects."
...
+    "crates/bender-slang",
+    "crates/bender-kg-models",
+    "crates/bender-kg-extract",
+    "crates/bender-kg-store",
+    "crates/bender-kg-similarity",
+    "crates/bender-kg-core",
+    "crates/bender-kg-mcp",
+]
+
+[workspace.dependencies]
```
micprog, line 28: Is there a way to move these dependencies to the corresponding packages, or can we gate them only for when required? Just wondering regarding clutter.

src/sess.rs
```diff
@@ -453,7 +453,10 @@ impl<'ctx> Session<'ctx> {
             .map(|file| match *file {
                 config::SourceFile::File(ref path) => {
                     let ty = match path.extension().and_then(std::ffi::OsStr::to_str) {
-                        Some("sv") | Some("v") | Some("vp") | Some("svh") => {
+                        // `.vh` headers are de-facto SystemVerilog macro files;
+                        // downstream tools (VCS, slang, verilator) parse them as part
+                        // of the unit, so classify them as Verilog here too.
+                        Some("sv") | Some("v") | Some("vp") | Some("svh") | Some("vh") => {
```
micprog, line 459: This is an unrelated change, if required move to a separate PR

src/cmd.rs
```diff
@@ -16,6 +16,8 @@ pub mod completion;
 pub mod config;
 pub mod fusesoc;
 pub mod init;
+#[cfg(feature = "slang")]
+pub mod kg;
```
micprog, line 20: As mentioned by the Cargo.toml, I'd move this to a separate new feature, not the slang feature. This is repeated across several files, I will not explicitly mention everywhere.

micprog (commented):
This looks like a very large change/addition, created with a lot of AI support. I added some initial feedback to encourage separatation of concerns, but I'm sceptical if this can be reasonably integrated. As I see the use of having access to bender's internal information, I'm wondering if it would make more sense to push this into a separate project, extending bender itself to allow running of plugins that have access to some of the internal functions, e.g., WASM plugins with a bender-internal WASM runtime. I don't think we can confidently maintain such a large addition inside bender itself. What do you think?

Cargo.toml
aottavianoTT: Sure we can

src/sess.rs
aottavianoTT: Yes, I will actually review each diff in detail once we define an integration strategy to keep the PR in scope. However, good catch

src/cmd.rs
aottavianoTT: I gated it with `slang` because what this tool basically does is to map the outcome of a slang pass (the CST, or the typed AST if one also "elaborates" a single top to resolve param widths etc) into a graph on-disk that can be queried, so in my taxonomy they are closely related. 

However, I would agree that any fronted that can produce an AST would also work, since the graph is populated from a frontend-agnostic data structure. If we make this a separate plugin, we probably need to keep tool-agnostic interfaces as it is right now, but this conditional dependency problem in Cargo would be solved (i.e., it is now a "separate" tool/crate)

aottavianoTT:
> This looks like a very large change/addition, created with a lot of AI support. I added some initial feedback to encourage separatation of concerns, but I'm sceptical if this can be reasonably integrated. As I see the use of having access to bender's internal information, I'm wondering if it would make more sense to push this into a separate project, extending bender itself to allow running of plugins that have access to some of the internal functions, e.g., WASM plugins with a bender-internal WASM runtime. I don't think we can confidently maintain such a large addition inside bender itself. What do you think?

Hi @micprog, thanks for the comments and initial review. I agree that this feature is not a cheap add-on, although I tried to re-use as much as what bender was offering already and only conditionally add differentiating features in a tool-agnostic way (basically, mapping the simple CST or the typed AST from slang to a generic data structure that could be generated from other fronteds, and then upsert an on-disk graph database from this data structure).

I am fine with having this live in a different repo, however I understand this plug-in idea does not exist yet in Bender, or is there a reference example already? If not, we can build it by leveraging this use case, making sure it is generic enough to work with other examples, and then integrate this use case itself.

What do you think? Can draft a proposal and we can iterate on it. I'll probably re-use this PR

PR #307, opened 2026-05-28
Add git database locks and db_dir config

micprog:
- **Filesystem locks**: wrap `git_database` and `checkout_git` in a cross-process `flock` (`fs4`) keyed per dependency, so concurrent `bender` invocations against the same dep serialize safely. Locks live at `<database>/git/locks/` and are released by the kernel on process exit.
- **`db_dir` config field**: new optional setting that overrides only the bare-repo and lock paths, leaving checkouts under `database`. Lets a single shared `Bender.local` enable cross-project cache reuse without forcing per-project `workspace.checkout_dir`. Older bender versions silently ignore the field and fall back to per-project caches — safer than the legacy `database:`-only sharing recipe in mixed-version setups.
- **Clippy cleanup**: separate commit fixing pre-existing warnings (`collapsible_match`, redundant `.into_iter()`, `args(&[...])`). `cargo clippy --all-targets -- -D warnings` now passes.

Docs updated: `configuration.md` (new `db_dir` entry), `workflow/ci.md` (`db_dir` is now the recommended sharing recipe; retires the concurrent-runs caveat), `local.md`, `CHANGELOG.md`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,7 @@ indicatif = "0.18.3"
 regex = "1.12.2"
 log = "0.4"
 env_logger = { version = "0.11", default-features = false, features = ["auto-color"] }
+fs4 = "1.1"
 
 [target.'cfg(windows)'.dependencies]
 dunce = "1.0.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ description = "A dependency management tool for hardware projects."
 readme = "README.md"
 license = "Apache-2.0 OR MIT"
 edition = "2024"
-rust-version = "1.87.0"
+rust-version = "1.89.0"
 
 [workspace]
 members = ["crates/bender-slang"]
@@ -53,7 +53,6 @@ indicatif = "0.18.3"
 regex = "1.12.2"
 log = "0.4"
 env_logger = { version = "0.11", default-features = false, features = ["auto-color"] }
-fs4 = "1.1"
 
 [target.'cfg(windows)'.dependencies]
 dunce = "1.0.4"
--- a/book/src/local.md
+++ b/book/src/local.md
@@ -42,7 +42,7 @@ database: my_deps_cache
 
 # Share only the bare git repos and lock files across projects, while
 # keeping per-project checkouts under each project's own .bender/.
-# Older Bender versions silently ignore this field.
+# Bender versions before 0.32 silently ignore this field.
 db_dir: /var/cache/bender_shared
 
 # Use a custom git binary or wrapper
--- a/book/src/local.md
+++ b/book/src/local.md
@@ -40,6 +40,11 @@ For a detailed guide on using these commands for multi-package development, see 
 # Change the directory where dependencies are stored (default is .bender)
 database: my_deps_cache
 
+# Share only the bare git repos and lock files across projects, while
+# keeping per-project checkouts under each project's own .bender/.
+# Older Bender versions silently ignore this field.
+db_dir: /var/cache/bender_shared
+
 # Use a custom git binary or wrapper
 git: /usr/local/bin/git-wrapper.sh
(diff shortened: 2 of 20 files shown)


Cargo.toml
```diff
@@ -53,6 +53,7 @@ indicatif = "0.18.3"
 regex = "1.12.2"
 log = "0.4"
 env_logger = { version = "0.11", default-features = false, features = ["auto-color"] }
+fs4 = "1.1"
```
fischeti: The standard library also has locking capabilities since 1.89.0 (with the same API). Although this requires increasing the MSRV

book/src/local.md
```diff
@@ -40,6 +40,11 @@ For a detailed guide on using these commands for multi-package development, see
 # Change the directory where dependencies are stored (default is .bender)
 database: my_deps_cache
 
+# Share only the bare git repos and lock files across projects, while
+# keeping per-project checkouts under each project's own .bender/.
+# Older Bender versions silently ignore this field.
```
fischeti: maybe the exact version could be stated

book/src/workflow/ci.md
```diff
@@ -87,16 +87,26 @@ cache:
...
-This way, every job that runs in the runner reuses the already-fetched Git data instead of re-cloning from scratch. See [Configuration](../configuration.md) for more on the `database` setting.
+Every job on the runner now reuses the already-fetched Git data and serializes safely against concurrent jobs via per-dependency filesystem locks living next to the bare repos (`<db_dir>/git/locks/`).
+
+If you'd rather not place a `Bender.local` on the runner at all, exporting `BENDER_DB_DIR=/var/cache/bender_shared` in the job environment has the same effect — any project that explicitly sets `db_dir` in its own configuration overrides the env var.
+
+> **Older Bender versions** silently ignore both `db_dir` and `BENDER_DB_DIR` and fall back to their normal per-project `.bender/` cache, so the shared config above is safe to deploy on a runner that still hosts pinned-to-old-bender projects — those projects simply won't benefit from the shared cache, but they won't misbehave either.
+
+### Legacy `database:` recipe
+
+Earlier docs recommended sharing via [`database`](../configuration.md#database), which relocates both the bare repos *and* the checkouts:
```
fischeti: I don't think we need to reference "earlier" docs that have not been released yet

book/src/workflow/ci.md
```diff
@@ -87,16 +87,26 @@ cache:
...
-database: /var/cache/bender_shared
+db_dir: /var/cache/bender_shared
 ```
 
-This way, every job that runs in the runner reuses the already-fetched Git data instead of re-cloning from scratch. See [Configuration](../configuration.md) for more on the `database` setting.
+Every job on the runner now reuses the already-fetched Git data and serializes safely against concurrent jobs via per-dependency filesystem locks living next to the bare repos (`<db_dir>/git/locks/`).
+
+If you'd rather not place a `Bender.local` on the runner at all, exporting `BENDER_DB_DIR=/var/cache/bender_shared` in the job environment has the same effect — any project that explicitly sets `db_dir` in its own configuration overrides the env var.
+
+> **Older Bender versions** silently ignore both `db_dir` and `BENDER_DB_DIR` and fall back to their normal per-project `.bender/` cache, so the shared config above is safe to deploy on a runner that still hosts pinned-to-old-bender projects — those projects simply won't benefit from the shared cache, but they won't misbehave either.
```
fischeti: maybe the exact version could be stated

fischeti (approved):
Nice addition! I definitely think sharing the database is the way to go. I have some optional comments, but otherwise LGTM.

Cargo.toml
micprog: Nice catch, MSRV bump drags with it a few clippy changes, but I think it makes more sense to bump that and align.

PR #308, opened 2026-05-29
Use shared clone for checkouts, dissociate on `bender clone`

micprog:
Clone git checkouts with `--shared` so they borrow objects from the bare git database via alternates instead of copying the full object store, making checkouts faster and smaller on disk.

Since the resulting checkout is no longer self-contained, `bender clone` now repacks the copied working directory (`git repack -a -d`) and removes the alternates link so the cloned-out repository is standalone.

### DIFF
--- a/src/cmd/clone.rs
+++ b/src/cmd/clone.rs
@@ -101,6 +101,27 @@ pub fn run(sess: &Session, path: &Path, args: &CloneArgs) -> Result<()> {
             if !command.unwrap().success() {
                 bail!("Copying {} failed", dep);
             }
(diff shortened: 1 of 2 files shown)


fischeti (approved):
I like it! Git is pretty smart in saving storage meaning local clones imply `--local` which will hardlink objects. This means cloning with `--shared` will mainly be useful if database and checkout are on different file systems.

The repack also makes sense to me. I was suprised that we do an actual copy of the repository and not another git clone, but I guess that's because the checkout can be dirty.

PR #309, opened 2026-06-02
Add support for unparseable/encrypted files

micprog:


fischeti:
The changes makes sense in general. I have some ideas on how to make the slang interface cleaner. Particularly the `*_indices` calls are not very elegant incl. the existing `reachable_tree_indices`. I implemented it like this previously, since `cxx` does not allow to pass vectors of syntax trees over the FFI e.g. `Vec<slang::syntax::SyntaxTree>`. But talking to the clanker, there might actually be a way around that which would result in a much more elegant FFI in general. I will have a go at it and will push some (maybe bigger) changes on top🤓

src/cmd/script.rs
```diff
@@ -103,6 +103,16 @@ pub struct ScriptArgs {
...
+    /// Drop files slang could not parse (default: keep them)
+    #[cfg(feature = "slang")]
+    #[arg(long, global = true, help_heading = "General Script Options")]
+    pub drop_unparseable: bool,
+
+    /// Tolerate slang parse errors in non-encrypted files instead of failing
+    #[cfg(feature = "slang")]
+    #[arg(long, global = true, help_heading = "General Script Options")]
+    pub allow_broken: bool,
+
```
fischeti: what's the difference between those two flags?

src/cmd/script.rs
```diff
@@ -103,6 +103,16 @@ pub struct ScriptArgs {
...
     pub trim_incdirs: TrimIncdirs,
 
+    /// Drop files slang could not parse (default: keep them)
+    #[cfg(feature = "slang")]
+    #[arg(long, global = true, help_heading = "General Script Options")]
+    pub drop_unparseable: bool,
+
+    /// Tolerate slang parse errors in non-encrypted files instead of failing
+    #[cfg(feature = "slang")]
+    #[arg(long, global = true, help_heading = "General Script Options")]
```
fischeti: potentially, it could make sense to group the slang flags separately

fischeti (commented):
I refactored the `bender-slang` FFI now which looks cleaner now imo. A new `ParsedTree` type now bundles the actual syntax tree as well as `parsed_ok` and `encrypted` flags that can be probed easily. I also bumped the `bender-slang` version again.

The only remaining issue I have with it now are the `--drop-unparseable` and `--allow-broken` flags. For me it is not entirely clear how they differ resp. what the actual use cases are.

src/cmd/script.rs
micprog: I refactored the flags for clarity - now we have new flags with three options each: `--broken [error]|keep|drop`, `--encrypted error|[keep]|drop`

src/cmd/script.rs
micprog: Done

fischeti (approved):
LGTM now. Thanks for the changes! I think the new flags are much more intuitive

PR #310, opened 2026-06-02
Force ssh batch mode to avoid hidden auth prompts during fetch

fischeti:
`GIT_TERMINAL_PROMPT=0` only suppresses git's own prompts. Over SSH, auth prompts (host key, password, passphrase) come from `ssh` itself and go straight to the terminal, where they hide behind the progress bars — so the fetch looks like it hangs (typically when no SSH key is configured).

Fix: append `-o BatchMode=yes` to `GIT_SSH_COMMAND` so ssh fails fast with a clear error instead of prompting. An existing user `GIT_SSH_COMMAND` is preserved, and a user-set `BatchMode` still wins.

### DIFF
--- a/src/diagnostic.rs
+++ b/src/diagnostic.rs
@@ -80,6 +80,15 @@ impl Diagnostics {
         *guard = multiprogress;
     }
 
+    /// Whether progress bars are currently being rendered.
+    ///
+    /// When this is true, interactive git/ssh prompts must be suppressed, since
+    /// they would be drawn over by the progress bars. When false (e.g. with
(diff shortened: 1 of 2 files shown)


micprog:
This looks useful to avoid issues! I'm wondering if we can keep the prompts (both for git and ssh) when progress bars are disabled, as well as adding a help message if a prompt would have been required, pointing to the non-progress flag, such that users can still move forward even if things don't work as expected.

micprog (approved):
LGTM :+1:

PR #311, opened 2026-06-05
ci: add Dependabot for Cargo and GitHub Actions

fischeti:
Sets up dependabot for Rust dependencies and Github Actions on a weekly basis. A 7-day cooldown is configured for security reasons.

Tested on my fork e.g. https://github.com/fischeti/bender/pull/2

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,28 @@
+version: 2
+updates:
+  # Rust dependencies (Cargo.toml + Cargo.lock)
+  - package-ecosystem: cargo
+    directory: "/"
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 10
+    cooldown:
+      default-days: 7
+    groups:
+      # One PR for routine patch/minor bumps to keep noise down.
+      cargo-minor-patch:
+        update-types:
+          - minor
+          - patch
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,28 @@
...
+  - package-ecosystem: cargo
+    directory: "/"
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 10
+    cooldown:
+      default-days: 7
+    groups:
+      # One PR for routine patch/minor bumps to keep noise down.
+      cargo-minor-patch:
```
micprog, line 13: Does the bot also flag major bumps?

micprog (commented):
LGTM, ideally we also get notified of major bumps that we should align to - if it requires manual updates in the rest of the code we should then do that.

.github/dependabot.yml
fischeti: I had to double check, but yes! By default every kind of version bump results in a separate PR i.e. one PR per dependency. With the `groups` configuration, we just bundle all `minor` and `patch` into one PR to reduce noise, since those should usually not cause any problems.

micprog (approved):

Issue #312, opened 2026-06-08
bender script emits .svh header files as compile sources (regression in 0.32.0)


fischeti:
## Summary

Since `v0.32.0`, `bender script` emits SystemVerilog header files (`.svh`) that are listed in a package's `files:` as **standalone compile sources**. Previously (`v0.31.0`), such files were recognized as headers of unknown compile type (`W28`), kept **out** of the emitted source list, and reachable only via `+incdir+`.

This is a breaking change for the common convention of listing `.svh` headers in `files:` (so bender tracks them) while including them via `include_dirs`. Compiling a header standalone *and* via `` `include `` causes duplicate definitions — e.g. VCS reports class redefinition and fails elaboration.

## Minimal reproduction

`Bender.yml`:
```yaml
package:
  name: repro
  authors: ["repro"]

sources:
  - include_dirs:
      - inc
    files:
      - src/tracer.sv       # compiles; `include "trace_item.svh"
      - inc/trace_item.svh  # header: listed in files: for tracking, also on +incdir+inc
```
`src/tracer.sv`:
```systemverilog
`include "trace_item.svh"
module tracer; endmodule
```
`inc/trace_item.svh`:
```systemverilog
class trace_item; endclass
```

Run `bender script vcs` with each version:

**0.31.0** — header excluded from sources, warned as unknown type:
```
    "+incdir+$ROOT/inc" \
    "$ROOT/src/tracer.sv"
```
```
warning[W28]: File with unknown type:
  - /tmp/bender-svh-repro/inc/trace_item.svh
```

**0.32.0** — header now emitted as a compile source:
```
    "+incdir+$ROOT/inc" \
    "$ROOT/src/tracer.sv" \
    "$ROOT/inc/trace_item.svh"
```

Diff of the emitted source list (0.31.0 -> 0.32.0):
```diff
     "$ROOT/src/tracer.sv"
+    "$ROOT/inc/trace_item.svh"
```

## Real-world impact

cva6 lists its tracer headers in `files:` (under `target: not(synthesis)`), alongside the `.sv` that includes them:
```yaml
- include_dirs: [core/include, common/local/util]
  files:
    - common/local/util/instr_tracer.sv
    - common/local/util/instr_trace_item.svh
    - common/local/util/ex_trace_item.svh
```
With 0.32.0, `ex_trace_item.svh` is both `` `include``d (via `+incdir+common/local/util`) and compiled standalone, so VCS errors:
```
[...]

PR #313, opened 2026-06-09
fix(sess): remove `.svh` files from verilog pattern list

fischeti:
Addresses https://github.com/pulp-platform/bender/issues/312 which was a regression in v0.32.0

This fixes the issues with cva6 that I had in gwaihir


micprog (approved):
LGTM :+1:

We may need to address another item that would ensure all files, even unknown formats, are properly output to flist scripts, just to ensure that existing workflows with older requirements still work, but that's a separate PR.

PR #314, opened 2026-06-09
Add `git_submodules` flag to control dependency submodule cloning

fischeti:
## What

Adds a `git_submodules` option to control whether Bender clones the Git submodules of dependencies.

- **Config field** `git_submodules` (default `true`), mirroring the existing `git_lfs` option — OR-merged across config layers, validated with `unwrap_or(true)`.
- **CLI flag** `--git-submodules <true|false>` (env `BENDER_GIT_SUBMODULES`), modelled on `--git-throttle`. It overrides the configured value in **either** direction.
- When disabled, the `git submodule update --init --recursive` step is skipped during checkout (along with its per-submodule progress bar).

## Why

`git submodule update --init --recursive` currently runs unconditionally on every checkout that has a `.gitmodules` file, and it is frequently the slowest part of the initial dependency fetch. In top-level hardware projects, dependency submodules often hold software or tooling that is irrelevant to the build, so cloning them is wasted work. This gives users an escape hatch — persistently via config, or per-invocation via the flag (e.g. in CI).

The default stays `true`, so existing behavior is unchanged unless explicitly opted out.

## Notes

- The flag intentionally **takes a value** rather than being a presence-only switch (like `--no-progress`). This keeps the door open to later widening the accepted values — e.g. selecting specific dependencies — without a breaking change to the CLI surface. The config field stays a plain `bool` for now; no enum is introduced yet.
- Per-dependency / selective submodule control was deliberately *not* implemented here. The cleaner long-term home for that granularity is producer-side (a dependency declaring which submodule paths it needs), since the consumer generally lacks that knowledge and it doesn't scale to transitive deps.
- Docs updated: `book/src/configuration.md` (config field reference) and `book/src/dependencies.md` (Submodules section).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 
 ## Unreleased
 ### Added
-- Add `git_submodules` config field and `--git-submodules <true|false>` flag (env `BENDER_GIT_SUBMODULES`) to control cloning of dependency submodules; defaults to `true`, the flag overrides the configured value in either direction (https://github.com/pulp-platform/bender/pull/PRNUM).
+- Add `git_submodules` config field and `--git-submodules <true|false>` flag (env `BENDER_GIT_SUBMODULES`) to control cloning of dependency submodules; defaults to `true`, the flag overrides the configured value in either direction (https://github.com/pulp-platform/bender/pull/314).
 
 ## 0.32.0 - 2026-06-05
 ### Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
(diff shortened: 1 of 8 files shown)


micprog (commented):
I understand the frustration!

From separate conversations, we ideally add a configuration in the Bender.yml indicating which submodules are required, and which are not, on a per-dependency basis, possibly with additional target specification. In this instance, I like the approach to get at least something going. It would be great if we could have a command or option to force the checkout/update of all submodules in all dependencies, just in case we checked out without submodules and now need to get some/all. I think this is a good flag/subcommand for the checkout command, let me know what you think. To avoid breaking setups, I see an option to cleanly get all submodules as a prerequisite to this feature. Other than that, LGTM :+1:

fischeti:
I agree with the approach of specifying this on the side of the dependency, since the maintainer is more aware of which submodules are actually required. As discussed, the format would be something like:

```yaml
git_submodules:
  - sw/deps/printf
  - target: test
    submodule: sw/deps/cva6-sdk
    recursive: false # default: true
    shallow: false # default: true?
 # not updated if no specified explicitely
 # - pd/deps/ihp-130-pdk 
```

With this functionality in place, it would make sense to reverse the default to not clone i.e. `--git-submodule false` but keep it as a global switch (to avoid completely breaking setups).

I would tackle this in a separate PR. In that sense, this PR is ready as it is and it also does not include any breaking changes since the default stays the same and can be released as a patch version. Then, the following modifications would be released as a breaking release.

Let me know what you think!

micprog (approved):
I added a warning to help out in getting the submodule back in case the flag was inadvertently added, but now LGTM :+1:

fischeti:
> I added a warning to help out in getting the submodule back in case the flag was inadvertently added, but now LGTM 👍

LGTM!

PR #315, opened 2026-06-10
database: treat database as read-only after creation and use shared locks

fischeti:
In #307 we introduced database locks and separaring database from checkout in order to share the database as a cache cross projects. This is also relevant in the context of CI, since the git databases are only fetched once and can then be reused. However, the _exclusive_ locks on the databases essentially prevent any kind of parallelization and every invocation of bender becomes serialized across CI jobs.

This PR aims to relax the locks to use exclusive locks only when writing to the database, and use shared locks when reading from it.

The flow of fetching and checking out a git repository was slightly adapted to reduce the amount of write operations on a shared database:
* `git init` and `git fetch [<rev>]` are write operations and still acquire an exclusive lock
* The creation of a temporary tag `tmp-<hash>` in the database and the later checkout with `git clone --branch tmp-<hash>` has been replaced with `git clone --shared --no-checkout && git checkout <rev>`
* The database is configured to disable `git gc` since revisions might not be tracked anymore by refs i.e. the previous `tmp-<hash>` tags. This is maybe overly catious.

Furthermore, file locking is done more fine grained. Instead of locking the whole `git_database` call exclusively, only the part that requires writes (`git init` and `git fetch`) acquire an exclusive lock (if database is not ready yet).

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -28,7 +28,7 @@ The directory where Bender stores cloned and checked-out dependencies.
 - **Example:** `database: /var/cache/bender_dependencies`
 
 ### `db_dir`
-Optional override for the directory that holds bare git repositories and their lock files (i.e. `<db_dir>/git/db/` and `<db_dir>/git/locks/`). When set, it takes precedence over `database` (whether explicitly configured or left at its default) for these two paths only; the working-tree checkouts continue to follow `database` (or [`workspace.checkout_dir`](./manifest.md) in the project manifest). This makes it possible to share the heavy git data across projects on a persistent runner without also relocating per-project checkouts. See [Continuous Integration › Sharing the Database](./workflow/ci.md#sharing-the-database-across-runs-and-projects) for the recommended setup.
+Optional override for the directory that holds the bare git repositories and their lock files (both live under `<db_dir>/git/db/`, each lock file sitting next to the bare repo it guards). When set, it takes precedence over `database` (whether explicitly configured or left at its default) for this path only; the working-tree checkouts continue to follow `database` (or [`workspace.checkout_dir`](./manifest.md) in the project manifest). This makes it possible to share the heavy git data across projects on a persistent runner without also relocating per-project checkouts. See [Continuous Integration › Sharing the Database](./workflow/ci.md#sharing-the-database-across-runs-and-projects) for the recommended setup.
 - **Config Key:** `db_dir`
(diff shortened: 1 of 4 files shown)


fischeti:
Actually, I am not sure anymore if the locks are needed. I have the impression that git should be able to handle concurrency itself of a shared git database i.e. objects are immutable and do not need to be locked, while other things like refs seem to have fine-grained lock files:

https://stackoverflow.com/questions/19962024/locking-strategy-of-git-to-achieve-concurrency
https://stackoverflow.com/questions/750765/concurrency-in-a-git-repo-on-a-network-shared-folder#answer-751026

Edit: I guess the `git init` cannot be locked properly by git and could cause problems and apparently git also errors out when failing to acquire locks, so it could make sense after all🤓

micprog:
> Actually, I am not sure anymore if the locks are needed. I have the impression that git should be able to handle concurrency itself of a shared git database i.e. objects are immutable and do not need to be locked, while other things like refs seem to have fine-grained lock files:
> 
> https://stackoverflow.com/questions/19962024/locking-strategy-of-git-to-achieve-concurrency https://stackoverflow.com/questions/750765/concurrency-in-a-git-repo-on-a-network-shared-folder#answer-751026


I think we should preserve locks, and I like your solution to speed things up (still need to review the details). The main issue I'm attempting to solve is with multiple repos fetching the same bare ref. Concurrent fetches racing on the same ref would produce `fatal: cannot lock ref ...: File exists` from one of them. Git handles it cleanly but bender has no retry layer, so we'd just propagate that as a CI failure. Rare, but the failure mode is "this CI job randomly fails once a week". Keeping locks fixes this.

fischeti:
Yes, you are right🤓 I just realized that git locks are not blocking but just error out

micprog (approved):
I added some additional commits on top to add additional protections and improve reporting. Offline we discussed cleaning up some of the sloppy comments, other than that LGTM :+1:

fischeti:
I refactored the locks a bit, and removed excessive AI comments. Further, I aligned the location of the locks between database and checkout. The new structure looks like this:

```
<database>/
  └── git/
      ├── db/
      │   ├── <name>-<urlhash>/
      │   └── <name>-<urlhash>.lock
      └── checkouts/
          ├── <name>-<srchash>/
          └── <name>-<srchash>.lock
```

Let me know if that works for you, or whether you liked the explicit `git/locks` directory more.

PR #316, opened 2026-06-15
script: keep untyped files in flist output

micprog:
Regression from v0.30.0 (d0c1e13): switching the flist/flist-plus templates from `all_files` to the per-type split source groups silently dropped files whose extension isn't a recognized HDL type (.c, .tcl, .txt, ...), because untyped files were never emitted into those groups.

Emit untyped files as their own groups (file_type = None) so they reappear in flist/flist-plus. Tool-script formats that branch on file type (vsim, vcs, synopsys, formality, riviera, genus, precision) previously fell through their `file_type != ''` guard and emitted broken compile commands for such files; they now comment them out instead. Type-agnostic (vivado) and Verilog-only (verilator) formats are unaffected.

Add a regression fixture with a non-HDL file plus tests covering flist inclusion and tool-script commenting.

### DIFF
--- a/src/cmd/script.rs
+++ b/src/cmd/script.rs
@@ -900,8 +900,12 @@ fn emit_template(
         }
         separate_files_in_group(
             src,
+            // Categorize by file type. The extra `Some(..)` wrapper means untyped files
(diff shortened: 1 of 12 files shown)


fischeti (approved):
LGTM!

PR #327, opened 2026-07-03
book: Add `tb` target convention

colluca:
This PR suggests to split the conventional `test` target into separate `test` and `tb` targets. `test` includes reusable test sources that might be used by dependent repos. `tb` includes repo-specific testbenches that are not expected to be reused by other repos.

See https://github.com/pulp-platform/common_cells/pull/340 for a specific use case.


micprog (approved):
OK for me

PR #328, opened 2026-07-03
flake: fix build with slang feature

Smephite:
The `slang` feature (enabled by default) requires cmake and network access for FetchContent dependencies (slang, fmt, mimalloc), both unavailable in the nix sandbox. This breaks the flake build.

To support sandboxed builds without patching, `build.rs` now accepts optional env vars (`SLANG_SRC_DIR`, `FMT_SRC_DIR`, `MIMALLOC_SRC_DIR`). When set, they are forwarded as `FETCHCONTENT_SOURCE_DIR_*` cmake variables and used to derive include paths. Normal builds are unaffected — cmake still fetches everything when the vars are unset. This also simplifies the nixpkgs packaging in NixOS/nixpkgs#532944, which currently patches `build.rs` with `replaceVars` to achieve the same effect.

The flake is updated to use this mechanism: pre-fetched GitHub sources are passed via the env vars, `cmake` and `python3` are added as build inputs, source filters are extended for C++/test files, `NO_COLOR=1` is set for owo-colors test compatibility, and `run_all.sh` is run as an integration test.

### DIFF
--- a/crates/bender-slang/build.rs
+++ b/crates/bender-slang/build.rs
@@ -45,6 +45,13 @@ fn main() {
         vec!["-std=c++20"]
     };
 
+    // Allow overriding FetchContent source directories via environment variables.
+    // This is used by Nix (and other sandboxed build systems) where network access
+    // is blocked and dependencies must be pre-fetched.
+    let slang_src_dir = std::env::var("SLANG_SRC_DIR").ok();
+    let fmt_src_dir = std::env::var("FMT_SRC_DIR").ok();
+    let mimalloc_src_dir = std::env::var("MIMALLOC_SRC_DIR").ok();
+
     // Apply cmake configuration for Slang library
     slang_lib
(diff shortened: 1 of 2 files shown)


micprog (approved):
Looks OK to me, maybe @haflex wants to comment on the nix side?

fischeti (approved):
The changes in `build.rs` looks good to me. I don't have any experience for flake nix, but I trust you there :)

The question is whether this requires a new release of `bender-slang`. Does nix pull from crates.io? or does it just build from source?

Smephite:
Thanks guys!

> The question is whether this requires a new release of `bender-slang`. Does nix pull from crates.io? or does it just build from source?

No release needed. `bender-slang` is a path dependency in `Cargo.toml`, so nix builds it directly from the source tree. It doesn't pull individual crates from crates.io as pre-built binaries; all dependencies are fetched as source via Cargo.lock and compiled from scratch inside the nix sandbox.

PR #331, opened 2026-07-15
script: apply override_files before validation and slang

micprog:
Override files were resolved late, in the emit_template stage, so they did not apply to file-existence validation nor to the slang file-list reduction. This meant a shadowed original still had to exist on disk, and slang parsed both the original and its override as duplicate module definitions.

Resolve override_files right after flatten() instead: drop the override groups and rewrite like-named files in the regular groups to the overriding path (preserving SourceType). Only the overriding files are then validated, and slang analyzes/reduces them instead of the files they shadow. The OVERRIDDEN annotation is carried through a side map so --source-annotations output is unchanged.

### DIFF
--- a/src/cmd/script.rs
+++ b/src/cmd/script.rs
@@ -369,9 +369,15 @@ pub fn run(sess: &Session, args: &ScriptArgs) -> Result<()> {
         .filter_packages(&packages, args.keep_excluded_incdirs)
         .unwrap_or_default();
 
-    // Flatten and validate the sources.
+    // Flatten the sources, then resolve `override_files` groups before anything else inspects
+    // the file set. Doing this here — rather than at template-emit time — means the replaced
+    // files never reach validation or the slang pass: only the overriding files have their
+    // existence checked, and slang analyzes and reduces the overriding files instead of the
+    // ones they shadow (previously it saw both, colliding as duplicate module definitions).
+    let (srcs, override_comments) = apply_file_overrides(srcs.flatten());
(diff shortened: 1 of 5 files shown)


colluca (approved):
I've tested this in Gwaihir, on both a target that uses the Slang features and a target without and it seems to correctly fix the issues I was facing: E31 is no longer raised on the overridden file, and Slang also no longer fails with a missing file error.

Thanks again @micprog!

src/cmd/script.rs
```diff
@@ -808,31 +886,23 @@ fn emit_template(
...
     let mut all_defines = IndexMap::new();
     let mut all_incdirs = IndexSet::new();
-    let mut all_files = IndexSet::new();
+    let mut all_files: IndexSet<&Path> = IndexSet::new();
     let mut all_verilog = vec![];
     let mut all_vhdl = vec![];
     let mut unknown_files = vec![];
-    let mut all_override_files: IndexSet<(&Path, &str)> = IndexSet::new();
+    // `override_files` groups were already resolved in `apply_file_overrides`, so every group
+    // here contributes its files directly.
```
fischeti: Comment is a bit redundant

fischeti (approved):
Makes sense and LGTM! just a minor optional comment

src/cmd/script.rs
micprog: Thanks! I removed the comment in the rebase :+1:

PR #333, opened 2026-07-16
clone: rename -p/--path flag to --working-dir

colluca:
## Summary
- `bender clone` used `-p`/`--path` and `bender snapshot` used `--working-dir` for the same working-directory concept (same default value `working_dir`, same semantics). `--working-dir` is now the primary, documented flag for `clone` too, matching `snapshot`.
- `-p`/`--path` are kept as fully-working hidden aliases on `clone` for backwards compatibility — neither appears in `--help`, but both still parse.
- Updated the `docs/book` references and added a CHANGELOG entry.

Not a breaking change: existing `-p`/`--path` usage keeps working.

## Test plan
- [x] `cargo check --no-default-features` passes
- [x] `bender clone --help` shows only `--working-dir <WORKING_DIR>` (no `-p`/`--path`)
- [x] `bender clone -p <dir> <pkg>` and `bender clone --path <dir> <pkg>` still parse correctly
- [x] `bender snapshot --help` shows `--working-dir <WORKING_DIR>` (unchanged)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 
 ## Unreleased
 ### Changed
-- `bender clone`: rename `-p`/`--path` flag to `--working-dir` to match `bender snapshot`'s flag for the same concept.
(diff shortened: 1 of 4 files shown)


fischeti (approved):
LGTM! thanks @colluca

micprog (approved):
LGTM, thanks! :+1:

PR #334, opened 2026-07-17
script: expose slang-resolved headers as `all_headers` for templates

colluca:
## Summary

Adds a new `all_headers` variable to the `bender script` template context: the
fine-grained, transitive, deduplicated set of header files slang actually
resolved via `` `include `` while parsing, mirroring what `verilator -MMD`
reports.

Today `flist` emits only source files and `flist-plus` adds `+incdir+`
*directories* — neither surfaces the individual header files that were pulled
in. slang already tracks them (`resolved_include_paths`), and
`apply_slang_filters` already computed this list for `--trim-incdirs`, then
discarded it. This PR keeps that list and exposes it to templates, so users can
build precise dependency lists (e.g. a Makefile `.d`) that invalidate correctly
when a header changes.

## What changed

- `apply_slang_filters` now returns the resolved-header paths (computed
  whenever the slang pass runs, instead of only for `--trim-incdirs`).
- `run()` threads them through to `emit_template`, which inserts a sorted,
  deduplicated `all_headers` (absolute paths, consistent with
  `all_files`/`all_incdirs`) into the Tera context.
- `all_headers` is populated whenever the slang pass runs (`--top`,
  `--trim-incdirs {always | auto+--top}`, or a `--broken`/`--encrypted`
  policy) and is empty otherwise.

No new C++ (the slang machinery already existed). **`flist`/`flist-plus` output
is unchanged** — the new variable is only consumed by templates
that reference it.

### DIFF
--- a/src/cmd/script.rs
+++ b/src/cmd/script.rs
@@ -380,7 +380,7 @@ pub fn run(sess: &Session, args: &ScriptArgs) -> Result<()> {
     // unused include directories (per `--trim-incdirs`), with per-class policies for files
     // slang couldn't fully parse (`--broken`, `--encrypted`).
     #[cfg(feature = "slang")]
-    let (srcs, unparseable_paths) = {
+    let (srcs, unparseable_paths, resolved_headers) = {
         let trim_incdirs = match args.trim_incdirs {
             TrimIncdirs::Always => true,
             TrimIncdirs::Never => false,
@@ -400,7 +400,11 @@ pub fn run(sess: &Session, args: &ScriptArgs) -> Result<()> {
             Some(ParsePolicy::Error) | Some(ParsePolicy::Drop)
         );
         if args.top.is_empty() && !trim_incdirs && !policies_need_slang {
-            (srcs, std::collections::HashSet::<PathBuf>::new())
+            (
+                srcs,
+                std::collections::HashSet::<PathBuf>::new(),
+                Vec::new(),
+            )
         } else {
             apply_slang_filters(
                 srcs,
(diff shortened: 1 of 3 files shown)


src/cmd/script.rs
```diff
@@ -852,6 +870,15 @@ fn emit_template(
     };
     tera_context.insert("all_incdirs", &all_incdirs);
 
+    // Fine-grained header files slang resolved via `include (transitively) for the kept trees.
+    // Empty unless the slang pass ran (i.e. `--top`/`--trim-incdirs`/a parse policy triggered it).
+    // Exposed so `template`/`template-json` can emit dependency lists (e.g. a Makefile `.d`) that
+    // track header edits precisely, without listing whole include dirs. Sorted+deduped; paths are
+    // absolute, consistent with `all_files`/`all_incdirs` (relativize via `root` in templates).
+    let mut all_headers: IndexSet<PathBuf> = resolved_headers.iter().cloned().collect();
+    all_headers.sort();
```
fischeti: not sure why this needs to be sorted (again)

fischeti (approved):
LGTM in general

src/cmd/script.rs
colluca: Good catch, this is redundant now. `resolved_headers` is already sorted before it gets threaded through here, and collecting into `IndexSet` only deduplicates while preserving that order. I will drop this extra sort.

micprog:
I'm not entirely sure on the sort that you now removed, it may be necessary, but I haven't looked into it enough. I think we can merge this now, it might be worth it to open an issue to get include directory sorting correct and reproducible and respecting dependency order, I think currently the directories also in all_incdirs may just be sorted alphabetically. Please rebase to resolve conflicts and I'll merge as-is for now :+1:

colluca:
@micprog rebased now :+1:

PR #337, opened 2026-07-31
chore(script)!: upgrade templating engine from Tera 1 to Tera 2

fischeti:
Supersedes #329. The dependabot bump does not compile: Tera 2.0 breaks both the Rust API and the template syntax, and all eleven built-in script templates fail to render against it.

## Rust API breakage

- **`Context::into_json()` is gone.** Tera 2's `Context` is effectively write-only — it derives only `Debug, Clone, PartialEq, Default`, wraps a `BTreeMap<Cow<str>, tera::Value>` with `data` as `pub(crate)`, and exposes no iterator. There is no way to get the values back out, which is exactly what `bender script template-json` needs. The values are now accumulated in a `serde_json::Map` and converted with `Context::from_serialize` immediately before rendering.
- **`render_str` gained an `autoescape: bool` argument**, since a one-off template has no filename to infer escaping from.

## Template breakage

Every built-in format failed to render. Three classes of change:

| Tera 1 | Tera 2 | Affected |
| --- | --- | --- |
| `define.0` / `define.1` | `define[0]` / `define[1]` | 10 templates — dot notation no longer indexes arrays |
| `is starting_with(root)` | `is starting_with(pat=root)` | `flist`, `flist-plus` — tests take keyword arguments |
| `trim_start_matches(pat='/')` | `trim_start(pat='/')` | `flist`, `flist-plus` — trim filters consolidated |

## Beyond the straight port

**Templates are registered by name and rendered with `render`** instead of going through `render_str` as a one-off. Errors now name the actual file:

```
error: Unknown filter `bogus_filter`
 --> /path/to/my_format.tera:2:50
  |
2 | {% for f in all_files %}{{ f.file }} {{ f.nope | bogus_filter }}
  |                                                  ^^^^^^^^^^^^
```

instead of `__tera_one_off:2:50`, which was useless for anyone using `--template`. Registering the template also runs Tera 2's reference validation, so unknown filters and tests surface as a parse error before any rendering happens. Autoescaping is disabled explicitly rather than left to the filename-suffix default, so a user template that happens to be named `.html` cannot silently start HTML-escaping its own output.

**`preserve_order` is enabled.** Tera 2 ships `default = []`, and its `Map` type is a `std::collections::HashMap` unless that feature is on — so map iteration order is randomized per process. Five consecutive runs of a template doing `{% for k, v in global_defines %}`:

```
ZULU=1 MIKE=3 ALPHA=2
ALPHA=2 ZULU=1 MIKE=3
ALPHA=2 MIKE=3 ZULU=1
MIKE=3 ALPHA=2 ZULU=1
ALPHA=2 ZULU=1 MIKE=3
```

Tera 1 used a `BTreeMap` and was deterministic. No built-in template iterates a map — they use `all_defines`, an array of pairs — so this only affects custom templates, but non-reproducible build scripts are a bad failure mode and an unpleasant one to debug. The feature switches Tera's map to `indexmap`, already a direct dependency of bender, so it adds nothing to the tree.

## Dependencies

Drops **44 transitive dependencies, adds none** (`Cargo.lock` −429 lines): `chrono`, `chrono-tz(-build)`, `rand(_chacha,_core)`, `pest` ×4, `globwalk`, `globset`, `ignore`, `phf` ×4, `slug`, `deunicode`, `humansize`, `percent-encoding`, `crossbeam-*` ×3, `zerocopy(-derive)`, five `windows-*` crates, and more. Tera 2 moved built-ins needing heavy dependencies into a separate opt-in `tera-contrib` crate.

## Verification

- **Output is byte-identical to Tera 1** across all 12 formats × {default, `--compilation-mode separate --source-annotations`, `-D FOO=1 -D BAR`}, plus `--only-defines`, `--only-includes` and `--relative-path` — 41 outputs, zero differences.
- `cargo test` passes (15 + 8 + 18), `cargo fmt --check` clean, no new clippy warnings.

## Notes for reviewers

- Resolves to **tera 2.1.0**, not the 2.0.0 in #329.
- `ScriptTemplate` bundles the template name and body into one parameter; adding a separate argument would have pushed `emit_template` to 8 and tripped `clippy::too_many_arguments`.
- **Pre-existing bug, not a regression:** `bender script vivado --compilation-mode common --only-defines` fails on undefined `all_files`, because that variable is only inserted when `emit_sources` is set. It fails the same way on `master` under Tera 1. Worth a separate fix — inserting the empty collections unconditionally.
- `{% include %}` in a custom template does not work, since only the single named template is registered. It failed under Tera 1 too. Supporting it would mean loading a template directory (`Tera::load_from_glob`, gated behind the `glob_fs` feature) — out of scope here.
- **Suggested follow-up:** a test that feeds the `include_str!` templates through `add_raw_templates` and asserts they all parse and validate. Every one of the eleven was broken by this bump and nothing caught it; each break only appeared when that specific format was run by hand.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Breaking Changes
+- script: Upgrade the templating engine from Tera 1 to Tera 2. Custom templates passed to `bender script template --template` may need updating. See the [Tera migration guide](https://github.com/Keats/tera/blob/master/MIGRATION.md) for the full list. The output of all built-in formats is unchanged.
+
 ### Added
 - script: expose `all_headers` to templates — the fine-grained, transitive, deduplicated set of header files Slang resolved via `` `include `` for the kept trees. Populated whenever the Slang pass runs (`--top`/`--trim-incdirs`/`--broken`/`--encrypted`).
 
 ### Changed
 - `bender clone`: primary flag for the checkout directory is now `--working-dir`, matching `bender snapshot`'s flag for the same concept; `-p`/`--path` are kept as hidden aliases for backwards compatibility.
+- script: Template errors now report the template name and source location (e.g. `my_format.tera:12:54`), and unknown filters or tests are reported before rendering starts.
 
 ### Fixed
 - script: apply `override_files` before validation and the slang pass, so overriding files replace their targets in file-existence checks and in `--top`/`--trim-incdirs` reduction (previously slang saw both the original and the override as duplicate modules); the overridden-file annotation is preserved.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,20 +4,11 @@ version = 4
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.4"
+version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
+checksum = "c982642fa9e8606056828ee9a8505737230110bb1099153c79efe865c59d12ba"
 dependencies = [
  "memchr",
-]
-
-[[package]]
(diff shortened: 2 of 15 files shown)


micprog (approved):
LGTM, thanks for pursuing this update! :+1:

It would be great if you could open a separate PR with fixes to the issues highlighted, even as they are preexisting.

PR #339, opened 2026-08-01
fix(rewriter): preserve scoped nested renames

### LINKED ISSUE #338: `bender pickle` does not properly rename scoped name in packed dimensions
`bender pickle` does not properly rename all the occurrences of a package, in particular we found the following minimal usecase that fails pickling.

[`Bender.yml`](https://drive.google.com/file/d/1NRbPqcUkdV9YSls5xN15DSKEvVlYd3_a/view?usp=sharing)
[`prefix_bug.sv`](https://drive.google.com/file/d/15TROkUdHX-zEAQKJyeW0cfpZcaqjfuNT/view?usp=sharing)

delivers wrong output when running `bender pickle --prefix pfx_ --expand-macros --top dut > out.sv`.

fischeti:
Fixes https://github.com/pulp-platform/bender/issues/338

Rewritting recursively with slang is a bit difficult with slang, since it does not persist for nested scoped names when replacing the syntax node. There is a helper function`rewrite_scoped_names_inplace` which solves this, but it was simply missing for this particular case. I added some regression tests as well.

### DIFF
--- a/crates/bender-slang/cpp/rewriter.cpp
+++ b/crates/bender-slang/cpp/rewriter.cpp
@@ -198,6 +198,8 @@ class ReferenceRewriter : public SyntaxRewriter<ReferenceRewriter> {
         ScopedNameSyntax* newNode = deepClone(node, alloc);
         newNode->left = newLeft;
 
+        rewrite_scoped_names_inplace(*newNode->right);
+
         replace(node, *newNode);
         refRenamed++;
     }
--- a/tests/pickle.rs
+++ b/tests/pickle.rs
@@ -147,6 +147,16 @@ mod tests {
     }
 
     #[test]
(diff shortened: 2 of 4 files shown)


thommythomaso (approved):

fischeti:
Superseded by #342, which fixes this at the root instead of adding another manual re-walk.

The underlying issue is that `replace()` makes the replaced subtree opaque to the clone walk, so `rewrite_scoped_names_inplace` has to be remembered at every replacement site — this PR is what happens when it isn't, and `VirtualInterfaceTypeSyntax` and `PackageImportItemSyntax` are two more sites on master that are still missing it. #342 switches every rename to a token-level edit (keyed by owning node + child index, applied during the clone walk), so nested names are picked up by the ordinary visitor traversal and the helper goes away entirely.

PR #342, opened 2026-08-03
fix(rewriter): rewrite renames as token edits instead of node replacements

### LINKED ISSUE #338: `bender pickle` does not properly rename scoped name in packed dimensions

fischeti:
Fixes #338. Replaces #339 — same test, different fix.

## What was wrong

The pickler only ever renames single identifier tokens, but it did so by rebuilding and replacing whole syntax nodes. In slang, replacing a node makes everything inside it invisible to the rewriter, so renames queued deeper in that subtree are silently dropped.

The workaround was `rewrite_scoped_names_inplace`, a manual re-walk that has to be added by hand at every place we replace a node. It's easy to forget, and #338 is a case where it was: in `common_pkg::state_t [common_pkg::NumStates-1:0]`, the packed dimension sits in the part of the tree that got skipped. Two other handlers on master (virtual interface types and package imports) are missing the same re-walk and have the same latent bug.

#339 adds one more manual re-walk. This PR removes the need for them.

## The fix

Slang also lets you replace a single *token* instead of a node. Token edits don't hide anything, so they combine freely with renames elsewhere in the tree. Since every rename we do is one token, no handler needs to replace a node at all.

Each handler now edits its token and then keeps walking as usual, so nested names get renamed by the normal traversal. `rewrite_scoped_names_inplace` and every `deepClone` are gone (-107/+62 in `rewriter.cpp`), and the virtual-interface and package-import cases are fixed along the way.

## Testing

- Test and fixtures from #339, unchanged. They pass here and fail against master, so they're covering the fix.
- All 9 pickle tests and 3 `bender-slang` tests pass.
- Pickling `tests/pickle` gives byte-identical output to master.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 
 ### Fixed
 - script: apply `override_files` before validation and the slang pass, so overriding files replace their targets in file-existence checks and in `--top`/`--trim-incdirs` reduction (previously slang saw both the original and the override as duplicate modules); the overridden-file annotation is preserved.
+- pickle: Rename scoped names nested inside a renamed scoped name, such as a packed dimension on a scoped type (`common_pkg::state_t [common_pkg::NumStates-1:0]`). The rewriter now applies renames as token edits rather than replacing whole syntax nodes, which also fixes the same class of missed rename in virtual interface types and package imports (https://github.com/pulp-platform/bender/pull/342).
 
 ## 0.32.1 - 2026-07-07
 ### Added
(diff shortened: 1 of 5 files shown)


micprog (approved):
LGTM, thanks for the fix! :+1:

PR #343, opened 2026-08-04
resolver: handle unbounded version requirements in conflict message

micprog:
Half-bounded or wildcard requirements (e.g. `>=1.0`, `*`) have no top and/or bottom bound, which panicked the unwraps when formatting the conflict-resolution message. Render the available bounds instead, and sort unbounded requirements as the extreme version in the sort helpers.


fischeti (approved):
LGTM!

PR #347, opened 2026-08-10
build: add x86_64 musl binaries

fischeti:
Adds Linux x86_64 binaries that link against musl instead of glibc. The advantage of that is that those are fully statically linked libraries so they are more portable e.g. to Linux distros with glibc versions lower than the manylinux container that we currently use or minimal containers without glibc at all.

The installer script will still install gnu based binaries if a compatible glibc is detected and fall back to musl builds otherwise.

I also tried adding ARM musl binaries, but it is much more complicated to set up with the current github runners (requires cross compilation and other things), which is why I decided to drop it.

## Verification

Release build passed in https://github.com/pulp-platform/bender/actions/runs/31378434929. I also tested the built artifact on my workstation with a couple of commands incl. slang based features.

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -32,6 +32,10 @@ jobs:
             runner: ubuntu-22.04-arm
             container: quay.io/pypa/manylinux_2_28_aarch64
             cache_golden_bin: false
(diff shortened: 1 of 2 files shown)


micprog (approved):
LGTM :+1:

PR #353, opened 2026-08-18
chore: override cargo-dist gh action versions

fischeti:
https://github.com/pulp-platform/bender/pull/323 fails CI because it tries to change the `release.yml` file that is generated by `cargo-dist`.

I found a setting for cargo-dist that allows to overwrite specific actions which should unblock the dependabot.


micprog (approved):
LGTM :+1:

PR #357, opened 2026-09-05
refactor(bender-slang)!: pass slices instead of vectors across FFI

fischeti:
Generally, all of those are anyway read-only and this gets rid of a few unnecessary re-allocations.


micprog (approved):
LGTM :+1:

micprog:
This will require a new bender-slang version to be released, right?

fischeti:
yes, I think it was anyway due, since there were some other changes as well.

PR #367, opened 2026-09-22
build(deps): remove obsolete/superseeded dependencies

fischeti:
`async-recursion`: async recursion is natively supported since Rust 1.77, `async-recursion` is not actively maintained anymore and pulls in old/duplicate `syn` version (although `miette` does as well currently).

`dirs`: can be replaced with `std::env`

`tokio`: the `full` target pulls in more than needed. I (Claude) selected the ones we are actually using, which removes some crates as well.


fischeti:
Closes https://github.com/pulp-platform/bender/pull/365 as well

